Issue #44, opened 2021-03-30
Create better layout for mobile/small screens


Janekdererste:
The current approach looks good on large window sizes but takes up too much space on small screens.

- The logo needs too much space (I think it takes up too much space in every case)
- Route alternatives are displayed as rows. This takes too much horizontal screen space
- The search box itself takes up too much space especially when more via points are displayed
- close or hide elevation widget

karussell:
For mobile I looked how others (qwant, google, ...) did this and I think the most important aspect is that the map should be as large as possible :D 

And e.g. when clicking on the input the rest of the panel can disappear (maybe even the map). The tricky thing is the alternatives and the vehicle options. I solved it here using an "expandable row" e.g. clicking on the "alternatives" shows all alternatives in one row below and similar for the vehicles and options. The alternatives could be even omitted (not sure if that is good UX) when we assume the user knows how to click on the "1 circle" or "2 circle" on the map. Additionally if the screen is still too small I put a fullscreen icon below or we could use full screen as soon as the map is dragged or something.

![image](https://user-images.githubusercontent.com/129644/112974012-3f489800-9152-11eb-805e-4e0d56b9dd63.png)

karussell:
Should we show the logo only when we type the location? (and/or remove the "maps" text of the logo? and/or move "GraphHopper" into the copyright text of the map?)

Janekdererste:
Elevatio Plugin should not be displayed on small sceens

karussell:
Related to #20

Janekdererste:
![grafik](https://user-images.githubusercontent.com/8005495/115233407-28c1ab00-a118-11eb-88ae-354e7764335f.png)
Compare different routing providers on mobile

Janekdererste:
![grafik](https://user-images.githubusercontent.com/8005495/115237753-375e9100-a11d-11eb-8381-5b675aa4d770.png)
Waze

Janekdererste:
![21-04-19 14-36-37 0217](https://user-images.githubusercontent.com/8005495/115238084-90c6c000-a11d-11eb-9c74-648f7f80dea4.png)
iOS (Apple Maps)

Janekdererste:
![21-04-19 14-52-28 0218](https://user-images.githubusercontent.com/8005495/115239923-87d6ee00-a11f-11eb-962e-4091f9db64f4.png)
Here maps (iOS)

Janekdererste:
[...]

PR #46, opened 2021-03-30
Add path details and elevation widget 

### LINKED ISSUE #12: info box with elevation, surface, max_speed

easbar:
Here I added the heightgraph we use for GH maps currently. It's a modified version of the original heightgraph that is currently living here: https://github.com/easbar/Leaflet.Heightgraph and can be used with both Leaflet and Mapbox (in contrast to the original which was only usable with Leaflet).

Fixes #12 

todo: 
- [x] svg icons from heightgraph are not loaded (need to adjust webpack config?!)
- [x] hide widget before path is calculated
- [ ] test (or hide?) on mobile


package.json
```diff
@@ -9,6 +9,7 @@
         "test": "jest"
     },
     "dependencies": {
+        "leaflet.heightgraph": "github:easbar/Leaflet.Heightgraph#61cd2e91935719caa93a9c73a6940c247e3212a0",
```
easbar, line 12: This should be ok for now, but in the longer run we should either clean up this repository so it can be used standalone or move the code we need to the maps repository.

src/routing/Api.ts
```diff
@@ -162,14 +163,17 @@ export async function info() {
...
         vehicle: args.vehicle || 'car',
-        elevation: false,
+        elevation: true,
         debug: false,
         instructions: true,
         locale: 'en',
         optimize: 'false',
-        points_encoded: true,
+        // todonow: apparently point decoding does not handle elevation yet
+        points_encoded: false,
```
easbar, line 172: Right? Shall we fix this first?

src/routing/Api.ts
```diff
@@ -162,14 +163,17 @@ export async function info() {
...
         'alternative_route.max_paths': 3,
         algorithm: 'alternative_route',
+        // todonow: simply hard-coded for now
+        details: ['road_environment', 'surface', 'average_speed'] as any,
```
easbar, line 176: Not sure yet how path details shall be chosen via the UI

easbar (commented):
For now the widget's state is managed within the widget (not using the react state handling). But this can be changed rather easily, because I already put all the map specific code into a separate file: https://github.com/easbar/Leaflet.Heightgraph/blob/mapbox/example/MapboxHeightGraph.js

src/routing/Api.ts
easbar: I just had to pass the is3D flag to a couple of methods, I think this is needed because we cannot tell from the encoded string whether it is 3D or not @karussell?

karussell: Yes, the response does not contain this information. (Should we include this to make it easier to consume on the client side?)

src/routing/Api.ts
karussell: This is fine until there is an options panel where we could put this (but not sure if this should be configurable at all)

(maybe add road_class and max_speed ?)

karussell:
>  test (or hide?) on mobile

Mobile for this widget is tricky (e.g. how to do this "hover mouse" functionality?) and so I would just hide it for now.

src/routing/Api.ts
easbar: No, I think it is easy enough to keep track of the request's elevation flag here. So unless there have been complaints about this in the past no need to change the response format IMO.

src/routing/Api.ts
easbar: I think sooner or later there definitely should be some way to configure this via the UI (at least some hidden/expert kind of option). But sure for now I will just add the other two encoded values you mentioned.

easbar:
> Mobile for this widget is tricky (e.g. how to do this "hover mouse" functionality?) and so I would just hide it for now.

Moving the mouse over the diagram to show the little flag at the corresponding route position indeed seems tricky for touch devices, but at least showing the elevation and path details would be useful I think. But maybe we can do this later :)

easbar:
@Janekdererste Can you take a look at this PR so we can merge it? And maybe you can help me to hide the widget for mobile? The css I pasted above does it, but I could not make it work with a `@media` query.

karussell:
Another option that we currently do is to "close" the widget on mobile per default:

image

but maybe this takes too much space?

easbar:
> but maybe this takes too much space?

Mobile could also be a tablet device, so it should depend on screen size. Also I think seeing the elevation details is quite useful especially for outdoor profiles that might be used from mobile devices primarily. I'd say let's hide it for now and later think about options to show it on mobile devices as well.

src/map/Mapbox.ts
```diff
@@ -83,6 +88,119 @@ export default class Mapbox {
...
+        );
+        const pathDetails = Object.entries(selectedPath.details).map(([detailName, details]) => {
+            const points = selectedPath.points.coordinates;
+            const features = [];
+            for (let i = 0; i < details.length; i++) {
+                let [from, to, value] = details[i];
+                if (typeof value === 'undefined' || value === null)
+                    value = 'Undefined'
+                features.push(this.createFeature(points.slice(from, to + 1), value));
+            }
```
Janekdererste, line 108: I'd shorten this like the following:
```
const features = details.map(([from, to, value = 'Undefined']: [number, number, any]) => 
                this.createFeature(points.slice(from, to + 1), value))
```
If value is always either string or not set, the last type of the triple could be string

src/map/Mapbox.ts
```diff
@@ -83,6 +88,119 @@ export default class Mapbox {
...
+            return this.createFeatureCollection(detailName, features);
+        });
+        const mappings: any = {};
+        mappings['Elevation [m]'] = function () {
+            return {text: 'Elevation [m]', color: '#27ce49'}
+        };
+        Object.entries(selectedPath.details).forEach(([detailName, details]) => {
```
Janekdererste, line 115: You could directly set the function property in the mapping directly like this:
```
const mappings: any = {
            'Elevation [m]': function () {
                return { text: 'Elevation [m]', color: '#27ce49' }
            },
        }
```

src/map/Mapbox.ts
```diff
@@ -83,6 +88,119 @@ export default class Mapbox {
...
+            mappings[detailName] = this.createColorMapping(details);
+        });
+        this.heightgraph.setData([elevation, ...pathDetails], mappings);
+    }
+
+    createFeature(coordinates: number[][], attributeType: number | string) {
```
Janekdererste, line 121: This method should be private. Like the other helper methods as well. I guess they could also be static since no state is used here.

Janekdererste (commented):
> 
> 
> > Mobile for this widget is tricky (e.g. how to do this "hover mouse" functionality?) and so I would just hide it for now.
> 
> Moving the mouse over the diagram to show the little flag at the corresponding route position indeed seems tricky for touch devices, but at least showing the elevation and path details would be useful I think. But maybe we can do this later :)
> 
> We could hide the widget using this css
> 
> ```
> .heightgraph {
>     display: none;
> }
> ```

Maybe it would be even better not to render it at all? For long routes the svg inside the control can become quite large. To not render it at all would give the best performance.  I don't know whether mapbox provides this but one could listen to a map resize event and if the map is too small (narrow) remove the control from the map? If such an event is not available one could listen for a window.resize event and hide or show the component based on a media query as defined in 

```Mapbox.ts: 306```

**Comment by Janekdererste** [comment]
> 
> 
> > test (or hide?) on mobile
> 
> Mobile for this widget is tricky (e.g. how to do this "hover mouse" functionality?) and so I would just hide it for now.

The Covid tracker on https://zeit.de has the exact same "hover" effect when the finger is moved on the diagram

File: src/map/Mapbox.ts [hunk 7]
```diff
@@ -6,6 +6,10 @@ import { SetPoint, SetSelectedPath } from '@/actions/Actions'
 import { Popup } from '@/map/Popup'
 import { FeatureCollection, LineString } from 'geojson'
 import { Bbox, Path } from '@/api/graphhopper'
+import mapboxgl from "mapbox-gl";
+window.mapboxgl = mapboxgl;
```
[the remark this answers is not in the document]
easbar: > Maybe it would be even better not to render it at all? For long routes the svg inside the control can become quite large.

Does the browser do rendering work even for when display is none?

easbar:
Thanks for the review and code suggestions! Let's postpone mobile-related questions to #44.

src/map/Mapbox.ts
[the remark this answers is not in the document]
Janekdererste: At least one can see the elements in the inspector. So, somehow they are part of the dom.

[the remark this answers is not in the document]
easbar: Ok then maybe we should use `user agent` or similar to determine whether or not the widget should be shown and do not add the map control in the first place.

PR #51, opened 2021-04-06
Faster mapload

karussell:
The initial map load should be faster. I tried disabling the animation vs. making it faster (see the 2 commits). Not sure what I like most. Without animation would be fine for me but maybe a fast animation is a nice visual gimmick?

Then I tried to get rid of the two map loads. The initial map load is done for `new Map()` and then the second is done due to:

> useEffect(() => map?.fitBounds(bbox), [bbox, map])

Can we avoid the first view somehow? IMO it is a strange experience that the map "flickers" on load.


Janekdererste:
I tried it and for my taste this is too fast. I'd prefer something between this animation speed and the default. 

On my machine the animation is basically skipped when the map is in fullscreen. I guess too many frames are skipped.

karussell:
Ok, sure. Feel free to modify this.

PR #52, opened 2021-04-06
Fix bbox not being set on routing result

### LINKED ISSUE #36: problem in fit bounds for identical points
http://gh-maps-react.s3-website.eu-central-1.amazonaws.com/?point=14.239884671170358%2C51.38756272791164&point=14.239884671170358%2C51.38756272791164&vehicle=car

Janekdererste:
fixes #36

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -32,15 +32,15 @@ export default function App() {
     })
 
     // only use the api info's bbox until any other bounding box was chosen. Is this too messy?
-    const chooseBoundingBox = function (infoBbox: Bbox, pathBbox: Bbox, shouldUseInfoBbox: boolean) {
-        if (shouldUseInfoBbox && pathBbox.every(num => num !== 0)) {
+    const chooseBoundingBox = function (infoBbox: Bbox, shouldUseInfoBbox: boolean, pathBbox?: Bbox) {
+        if (shouldUseInfoBbox && pathBbox && pathBbox.every(num => num !== 0)) {
             setUseInfoBbox(false)
             return pathBbox
         } else if (shouldUseInfoBbox) return infoBbox
-        return pathBbox
+        return pathBbox || [0, 0, 0, 0]
     }
 
-    const bbox = chooseBoundingBox(info.bbox, route.selectedPath.bbox, useInfoBbox)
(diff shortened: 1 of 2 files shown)


karussell (approved):
Thanks!

src/App.tsx
```diff
@@ -32,15 +32,15 @@ export default function App() {
...
     // only use the api info's bbox until any other bounding box was chosen. Is this too messy?
-    const chooseBoundingBox = function (infoBbox: Bbox, pathBbox: Bbox, shouldUseInfoBbox: boolean) {
-        if (shouldUseInfoBbox && pathBbox.every(num => num !== 0)) {
+    const chooseBoundingBox = function (infoBbox: Bbox, shouldUseInfoBbox: boolean, pathBbox?: Bbox) {
+        if (shouldUseInfoBbox && pathBbox && pathBbox.every(num => num !== 0)) {
             setUseInfoBbox(false)
             return pathBbox
         } else if (shouldUseInfoBbox) return infoBbox
-        return pathBbox
+        return pathBbox || [0, 0, 0, 0]
```
karussell, line 40: btw: we use this BBox as the "invalid case" here and elsewhere, but strictly speaking this is a perfect valid bbox :)

Still should not make a problem as this bbox is unrealistically small and points to nowhere

Janekdererste: The problem here was that the path coming from the api doesn't have a bbox property for routes with the same start and end point. Mapbox can actually handle the bbox with no area. 

The rest of the code relies on the bbox being at least present.

karussell: > The problem here was that the path coming from the api doesn't have a bbox property for routes with the same start and end point. 

Do you mean we include the bbox in the response of /route and do not include it when two coordinates are identical? Or do you mean with "api" our JS code?

karussell:
The problem is now that it zooms very close to the route even if it is very small and once it is the same points it zooms out to world wide. Instead we should probably define a minimum bbox **and** additionally ensure that the markers are inside this bbox.

src/App.tsx
karussell: Oh, indeed. This is a bug I think.

karussell: https://github.com/graphhopper/graphhopper/issues/2277

karussell:
https://github.com/graphhopper/graphhopper-maps/issues/53

PR #60, opened 2021-04-14
Automatically deploy maps

rokcarl:
This code will deploy the maps automatically. This will run a workflow in another repository that will do the actual deployment.

I've set up `PRIVATE_GITHUB_ACCESS_TOKEN` as a secret so this should work fine.

The deployment will be done for each merge/push to master. I've tested the functionality of this from my private test repository.


karussell:
Cool, thanks!

PR #63, opened 2021-04-17
single click on map should not set a marker

karussell:
This onClick behaviour is handy, but it often happens, especially on mobile, that the markers are changed due to accidentally clicking on the map

Should we revert the use of `useRef` as there is a comment:

> // use this to be able to use querypoints props in onClick callback of the map


Janekdererste:
> Should we revert the use of useRef as there is a comment:

The ref to the queryPoints should be removed as well. This was only to give access to the store's query points from within the closure of the onClick callback. As pointed out in the react documentation this construct is close to a nasty hack. So it would be good to get rid of it

PR #64, opened 2021-04-17
use maptiler instead of mapbox

karussell:


Janekdererste:
This looks all good in case you are waiting for approval :-)

PR #68, opened 2021-04-21
less visible precision for lat,lon in input

karussell:
Not sure if this is better done in a static utility method. As there are a couple of places where this is required (so I added the method to a class).


Janekdererste:
As you've mentioned, we've several places where a coordinate is transformed into text. There already is a (small) collection of converters here: https://github.com/graphhopper/graphhopper-maps/blob/master/src/Converters.ts I think it makes sense to have a ```coordinateToText``` according to ```metersToText``` using a ```new Intl.NumberFormat(undefined, { maximumFractionDigits: 6 })```

Then replace all asignments of queryText where the coordinate is used.

karussell:
Will do!

Issue #53, opened 2021-04-06
Improve bounding box for route


karussell:
If we have a very short route it is zoomed to it even if overview is lost. I.e. we need a minimum bbox.

Additionally ensure that the markers are inside this bbox.

Janekdererste:
The bounding box should have a minimum size and it should always include the markers

easbar:
Another problem is that the elevation widget might cover one of the query points, for example here:

https://graphhopper.com/maps2/?point=48.118447%2C11.539497&point=51.442113%2C14.265179&profile=car

So we should probably only take into account the available vertical space above the widget, and only use the entire vertical space when the widget is closed.

PR #71, opened 2021-04-25
avoid initial fitBounds

karussell:
this avoids the jump on load. the jump, when two points are specified, still exists.

TODO: zoom a bit into the world (if world boundary) to avoid seeing some continents twice.

related to #53

### DIFF
--- a/src/api/graphhopper.d.ts
+++ b/src/api/graphhopper.d.ts
@@ -1,5 +1,6 @@
 import { LineString } from 'geojson'
 
+// min lat, min lng, max lat, max lng
 export type Bbox = [number, number, number, number]
 
 export interface RoutingArgs {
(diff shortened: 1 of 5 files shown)


easbar:
Closing here in favor of #77. The initial jump when points are specified remains even in #77. I opened #82 for this.

src/api/graphhopper.d.ts
```diff
@@ -1,5 +1,6 @@
 import { LineString } from 'geojson'
 
+// min lat, min lng, max lat, max lng
```
easbar, line 3: I think it is actually min lng, min lat, max lng, max lat, for example world = [-180, -90, 180, 90], but yes, adding such a comment might help :)

karussell: oh, strange. I thought I did my research properly :/

karussell: Uh, looks like I had an inconsistent state in my mind ;)

https://github.com/graphhopper/graphhopper-maps/compare/master...test_current_location#diff-516d68fe9c8c9fc0d5cc075d0ba6348d77cee435c4b9eb0bad4c844d9af215b6R22

PR #74, opened 2021-05-14
#44 mobile layout

Janekdererste:
This PR attempts to solve #44 which is supposed to create a mobile friendly layout. The current status should be available here:

http://gh-maps-react.s3-website.eu-central-1.amazonaws.com/

@karussell and @easbar please have a look at the current version to provide some feedback. I assembled a list of things that need to be done before we can mark the issue as finished. Feel free to add items to that list by editing this comment if that's possible

What is still missing:
- [x] Context menu for touch input in the map panel
- [ ] ~~Selection of alternative routes. Currently one can select an alternative route by touching on the line displayed on the map. This is very hard to achieve though since the touch area is quite small.~~
- [x] the placeholder for alternative routes is still 3 routes even on mobile where only one is shown
- [ ] ~~Button styling for close button on full screen text search. Add "Esc"-button to leave this view too.~~
- [ ] ~~clear geocoding results when input loses focus~~
- [ ] ~~flip chevron of Query result to point upwards - maybe replace with something else entirely?~~
- [x] Show collapsed search view only if a route is available
- [x] attribution to OpenStreetMap etc needs to be done in mobile too. Add them to "Powered by" panel?
- [x] link should be www.graphhopper.com (currently it is just graphhopper.com)
- [ ] ~~scrolling instructions on mobile is somehow not possible in firefox~~ can't reproduce
- [ ] ~~selecting a different map on mobile nearly always results in selecting kurviger first before I can open it~~
- [x] collapsed search view breaks layout on firefox mobile. See comment and images below
- [x] graphhopper logo takes up too much space
- [x] Fix overflowing powered by container when autocomplete box is expanded

PR #75, opened 2021-05-19
Handle path details state within app

easbar:
todo:

- [x] should the path details be shown underneath or within the map? -> within the map
- [x] allow hiding the path details (should or should we not use the toggle icon?)
- [x] make the details responsive (resize when browser window size changes)
- [x] hide the graph when no path was calculated


src/App.module.css
```diff
@@ -5,7 +5,8 @@
     margin: 0;
     display: grid;
     grid-template-columns: 25rem auto auto;
-    grid-template-rows: 100%;
+    /* todo: adjust bottom pane height, make foldable and/or responsive */
+    grid-template-rows: auto 10rem;
```
easbar, line 9: @Janekdererste Can you help me here? I think we want to be able to hide/show the details using some button. Also the current css is not working very well as the details are moved to far down when we open the instructions list. For mobile I think we should hide the details entirely for now?!

src/pathDetails/PathDetails.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+export default function ({selectedPath}: PathDetailsProps) {
+    const containerRef: React.RefObject<HTMLDivElement> = useRef(null)
+    const [graph, setGraph] = useState<any | null>(null)
+    useEffect(() => {
+        const options = {
+            // todo: is this the right way to do this? for example it should adjust its size when the browser window
+            //       size is changed
+            width: containerRef.current!.clientWidth,
+            height: containerRef.current!.clientHeight,
```
easbar, line 22: This is another layout issue. I was not sure how to do this.

src/stores/PathDetailsStore.ts
```diff
@@ -0,0 +1,56 @@
...
+        if (action instanceof PathDetailsHover) {
+            return {
+                ...state,
+                pathDetailsPoint: action.pathDetailsPoint
+            }
+        } else if (action instanceof PathDetailsRangeSelected) {
+            return {
+                // todo: bug - when we change details and selected a range before and then zoomed somewhere else we jump
+                //       back to the box unintentionally
+                ...state,
```
easbar, line 43: This is still WIP..., but I opened the PR so maybe we can already fix some of the layout stuff

src/App.module.css
karussell: > For mobile I think we should hide the details entirely for now?!

I think this is done in https://github.com/graphhopper/graphhopper-maps/pull/74/files#diff-f5600dbc5cd82f7decd67da5f626d0205081d6a12e6da6ba1a5cbc14ea116c28R35

easbar: After the recent merge the heightgraph is no longer shown for small screens

src/stores/PathDetailsStore.ts
easbar: This was a problem in heightgraph and I fixed it there

src/pathDetails/PathDetails.tsx
```diff
@@ -0,0 +1,184 @@
...
+            //       them just for this...
+            expandControls: false,
+        }
+        const callbacks = {
+            pointSelectedCallback: onPathDetailHover,
+            areaSelectedCallback: onRangeSelected,
+            routeSegmentsSelectedCallback: onElevationSelected,
+        }
+        setGraph(new HeightGraph(containerRef.current, options, callbacks))
+    }, [containerRef])
```
Janekdererste, line 31: Is the old heightgraph properly removed from the dom? I.e. when the large and small layouts are exhchanged this component will be unmounted. I don't know from just looking at this. 

When the dependeny array is removed there are a lot of svg added to the dom which maybe indicate a missing cancel function as return value of this effect?

src/pathDetails/PathDetails.tsx
```diff
@@ -0,0 +1,184 @@
...
+    const resizeGraph = () => {
+        graph?.resize({ width: containerRef.current?.clientWidth, height: containerRef.current?.clientHeight })
+    }
```
Janekdererste, line 34: changed width to container's width

src/pathDetails/PathDetails.tsx
```diff
@@ -0,0 +1,184 @@
...
+    useEffect(() => {
+        window.addEventListener('resize', resizeGraph)
+        return () => window.removeEventListener('resize', resizeGraph)
+    })
+    useEffect(() => {
+        const pathDetailsData = buildPathDetailsData(selectedPath)
+        graph?.setData(pathDetailsData.data, pathDetailsData.mappings)
+    }, [selectedPath, graph])
+    return (
```
Janekdererste, line 43: It would be nice to hide the height graph if no route is selected. It doesn't look pleasant when empty something like:

```
return isPathPresent && 
        <div className={styles.layoutContainer}>
            <div className={styles.heightgraphContainer} ref={containerRef} />
        </div>
```

src/pathDetails/PathDetails.tsx
```diff
@@ -0,0 +1,184 @@
...
+    Object.entries(selectedPath.details).forEach(([detailName, details]) => {
+        mappings[detailName] = createColorMapping(details)
+    })
+    return {
+        data: [elevation, ...pathDetails],
+        mappings,
+    }
+}
+
+function createColorMapping(detail: any): any {
```
Janekdererste, line 93: I know you've ported it from js, but maybe remove the anys? Probably future you will be happy about it.

src/stores/PathDetailsStore.ts
```diff
@@ -0,0 +1,54 @@
...
+}
+
+export interface PathDetailsStoreState {
+    pathDetailsPoint: PathDetailsPoint | null
+    pathDetailBbox?: Bbox
+    pathDetailsHighlightedSegments: Coordinate[][]
+}
+
+export default class PathDetailsStore extends Store<PathDetailsStoreState> {
+    constructor() {
```
Janekdererste, line 20: The default constructor may be ommited

src/stores/PathDetailsStore.ts
```diff
@@ -0,0 +1,54 @@
...
+        super()
+    }
+
+    protected getInitialState(): PathDetailsStoreState {
+        return {
```
Janekdererste, line 25: I find it nice to not have null or undefined values. Other than Java Typescript somewhat forces us to never have undefined values since it is necessary to separately declare that type. Not having undefined and null removes a lot of places where one has to test for presence.

If you can make it work maybe it is possible to supply an empty but valid state initially and design the components so that they always display the right thing also with an initial state.

Janekdererste (commented):
Overall this looks good to me. I guess one can also see the limit of the used store pattern already. I think we are not there yet, but maybe soon.

src/stores/PathDetailsStore.ts
easbar: Yes, I also noticed this. Using types like `<type> | null` does seem a bit strange indeed. Then again it kind of expresses exactly what I want. `pathDetailsPoint` is the point on the map that corresponds to the mouse hover position in the height graph. When the mouse is not positioned above the heightgraph there is no such point and I use `null` to cover this state. What would be the alternative? We could use some special point value like `[600,600]` to represent 'no point selected', but so far I do not see why this would be better (and I think its worse tbh). Sure we would not need the extra `| null` in the typing but then we need to do another comparison like `point !== [600,600]` elsewhere. In this specific example there is only one place where we check if the point is `null` and this is the place where we have to decide whether there should or should not be a marker on the map. I think it would be the same if we had to check for some special value. 

But generally I agree that thinking about null or undefined values can make the code worse and my default assumption is also that nothing is null or undefined (same in Java btw) unless cases like this where there is a notion of a value possibly 'missing. Hm, another possibility could be adding something like `pointSelected: boolean` to the state?

src/pathDetails/PathDetails.tsx
easbar: Cool, thanks. Should the graph not use all available horizontal space? I thought this was kind of the original idea. But then it might 'collide' with the instructions list.. hm :)

src/pathDetails/PathDetails.tsx
easbar: Sure. This is somewhat related to our discussion about using `null` for missing values. The `selectedPath` is never null, but it might be missing. I would find it very intuitive to check if its null here :) Currently it might be an `RouteStore#emptyPath`, but to obtain the `isPathPresent` value here I need to figure this out by checking if the points of the path are empty or something like this.

easbar:
> Overall this looks good to me. I guess one can also see the limit of the used store pattern already. I think we are not there yet, but maybe soon.

Why do you think we are getting closer to this limit? There is quite a bit of boiler-plate code needed, but this PR does not really do anything we did not do elsewhere yet? But sure probably at some point we should think about our current approach again.

src/pathDetails/PathDetails.tsx
easbar: I added some types here even though I am not fully convinced this improved much :) But generally I agree that adding most or all the types will save time down the road. Here it is a bit awkward because we get these path details from the GH API which could either be `[number, number, number][]` or `[number, number, string][]` and there already is some code that needs to distinguish the two cases.

src/pathDetails/PathDetails.tsx
easbar: I tried this but could not make it work yet :( I also wonder if we should or should not show the toggle button as in current master. I think it is useful to have the button so one can hide the graph if needed. Or maybe it should just be resizable in general? Previously I thought we would put the graph underneath the map so I removed the icon.

src/pathDetails/PathDetails.tsx
easbar: The height graph is just a collection of <svg> elements attached to some parent <div>, I think. So when we remove the div the heightgraph is also gone?! But no idea tbh. I saw we can use some clean up function as return value of `useEffect`, but I am not sure what to remove?

src/pathDetails/PathDetails.tsx
Janekdererste: Yes, I gave it a container, which sits on top of the map. As it was with the widget before. It will not collide with the instruction list this way.

src/pathDetails/PathDetails.tsx
Janekdererste: I find a toggle button usefull.

src/stores/PathDetailsStore.ts
Janekdererste: I see that it makes sense in this case. I also agree that the special value approach would be worse.

Janekdererste:
> Why do you think we are getting closer to this limit?

Because we add more and more stores. Probably, this would be somthing a library would structure for us. 

Now, thinking about this comment a second time. Maybe having more and more stores could also indicate that it is an easy pattern to use and we"re still all good :-)

src/pathDetails/PathDetails.tsx
Janekdererste: It should be gone and I just noticed it when removing the dependencies as long as there is the containerRef dependency, I don't see more svg elements than  expected.

easbar: You mean there is nothing we need to do? :)

src/pathDetails/PathDetails.tsx
easbar: Ok I put back the toggle button, but I had to tweak the layout quite a bit for this. See my last commits

easbar: The graph is no hidden when there is no path

easbar:
> Probably, this would be somthing a library would structure for us.

Yes maybe at some point we should do a little comparison but currently I think our approach does the trick.

Can you please have a final look? From my side this is ready to merge.

src/pathDetails/PathDetails.tsx
Janekdererste: Yes, sorry about that comment. I typed it while thinking, not considering whether it would be helpful or not.

Janekdererste:
There is still a layout issue in firefox. I will look into it.

Janekdererste:
Okay, I solved the layout issues on firefox by making things simpler. 

I put the glue code to figure out the width of the graph back into the component. The component's container now forces the component to have the correct size. The width parameter which gets pushed down the heightgraph module is taken from that container. 

The height of the component's container is determined by the component itself I set it to 14rem which looks good on my screen and is slightly smaller than the default. I also limited the width to 1000px, which will make it span over all the available width on most screens I'd assume. 

The toggle button comes with its own background-color, corner-radius and drop-shadow. Maybe we can remove that. The code for this is not in our code base at the moment.

> Oh and maybe you want to adjust the fonts used in the path detail graph panel?

I didn't see an option to change it. If it is easily possible it would be great. Also to change the button as stated above.

I guess it is your tunr @easbar to have a final look now.

easbar:
Cool, thanks :+1: Your version works for me. I like that the graph resizing code is in `PathDetails.tsx` again. I think I could not make this work because I was missing this `overflow: hidden` rule or something (the container would not shrink). Also the close button is now shown at its proper position :) Did you remove the box shadow on purpose? The sidebar has it but the path details do not. 

> The height of the component's container is determined by the component itself I set it to 14rem which looks good on my screen and is slightly smaller than the default. I also limited the width to 1000px, which will make it span over all the available width on most screens I'd assume.

Sure

> I didn't see an option to change it. If it is easily possible it would be great. It would also be nice to style the toggle button accordinto our other controls.

Regarding the styles of the toggle icon and the fonts: They are defined in `node_modules/leaflet.heightgraph/src/heightgraph.css`. We can change this upstream to whatever we want or even ignore this file entirely and create out own version of it here, but can we not also overwrite these styles in our app somehow? If not, what would be the best way to make this more configurable upstream?

Janekdererste:
> Did you remove the box shadow on purpose

The box shadow wouldn't show when attached to the heightgraph directly. So I removed it because I wanted to finish. But it should have one. I can take another look at it. 

> Also the close button is now shown at its proper position

The button relies on the component's container to have ```position: relative```. If you want to use the heightgraph component somewhere else, it needs its own container which has that attribute. 

>  but can we not also overwrite these styles in our app somehow?

Probably that's the appropriate thing to do.

easbar:
Ok, but if these styling issues (box shadow, fonts etc.) aren't so easy to fix I'd say let's finish here now. For #87 I also had the problem that I did not manage to integrate the 'flag' that is shown on the map when we hover the path details and this might be related. And I am also not sure if we maybe should just move the entire diagram into our code base, so I think it's better to move on and postpone this for now unless there is a quick fix for the styling. I also do not understand why the font is now larger than in master since we did not really change it?

easbar:
I opened #105 for the remaining styling issues

PR #77, opened 2021-05-19
Improve handling of bbox state

### LINKED ISSUE #72: change map layer should not cause map bbox change
zoom in & change map layer => the new map layer has again world wide bbox instead of the previous

easbar:
Handling the bbox state (or more generally the view port state) is a bit tricky, because we partly control it from our application (using `/info`, `selectedPath.bbox` etc.) and partly it is maintained by direct user interactions with the map (zooming, scrolling etc.).

In some cases we just want to set the bounds of the map (`fitBounds`), for example when we want to zoom to a calculated route. This means we want to control the viewport ourselves. In other cases we want to keep the current viewport, for example when we change the map layer or clear a query input box. This means we want to neglect the bbox state of our application and the internal map state should take precedence. 

Here I refactored some of the bbox state handling and fixed #72. ~~Unfortunately with my current solution the map no longer zooms to the calculated route when the page is loaded initially. See more comments inline.~~ (<- this is fixed now)

Fixes: #72


src/App.tsx
```diff
@@ -36,16 +35,16 @@ export default function App() {
...
+    const [bbox, setBbox] = useState<Bbox>([-180, -90, 180, 90])
+    useEffect(() => {
+        // make sure the path bbox takes precedence over the info bbox
+        if (!route.selectedPath.bbox)
+            setBbox(info.bbox)
+    }, [info])
+    useEffect(() => {
+        if (route.selectedPath.bbox)
+            setBbox(route.selectedPath.bbox)
+    }, [route])
```
easbar, line 47: Using these two effects seems better than the previous code, because now we only update the bbox when the route or info changed, whereas previously we always had determine some bbox even when we did not want to change anything. This way we also no longer need the `shouldUseInfoBbox` flag.

src/map/Map.tsx
```diff
@@ -30,13 +33,14 @@ export default function({ selectedPath, paths, queryPoints, bbox, mapStyle }: Ma
                 Dispatcher.dispatch(new MapIsLoaded())
             }
         )
-        mapWrapper.fitBounds(bbox)
+        if (prevViewPort)
+            mapWrapper.setViewPort(prevViewPort)
```
easbar, line 37: This code is what was missing for #72. When we re-create the map instance we have to restore the previous viewport. However, just doing this is not enough, because without the changes in `App.tsx` there would be another call to `fitBounds` that sets the info or route bbox again.

src/map/Map.tsx
```diff
@@ -30,13 +33,14 @@ export default function({ selectedPath, paths, queryPoints, bbox, mapStyle }: Ma
...
         return () => map?.remove()
     }, [mapStyle])
     useEffect(() => map?.drawPaths(paths, selectedPath), [paths, selectedPath, map])
     useEffect(() => map?.showPathDetails(selectedPath), [selectedPath, map])
     useEffect(() => map?.drawMarkers(queryPoints), [queryPoints, map])
-    useEffect(() => map?.fitBounds(bbox), [bbox, map])
+    useEffect(() => map?.fitBounds(bbox), [bbox])
```
easbar, line 43: This seems to be the most tricky part to me. Keeping the `map` dependency means that after a style change the call to `setMap` will trigger `fitBounds` again. If we calculate a route, scroll somewhere else and then change the style this means the map will center at the calculated route again.

On the other hand if we do not keep the `map` dependency `map` will be undefined here when the page loads initially and the call to `fitBounds` won't happen. Not sure how to fix this tbh.

easbar: Update: With my last commit it is 'working' now: 2fadb177593d94e0f619888c5707b7d3c422973c, but honestly this seems way to complicated. I now use `useRef` to keep track of the latest bbox value so it is available by the time `onMapReady` is called and we can call `fitBounds(prevBbox.current)`. Is there maybe a better way to construct the map instance before we receive the bbox etc.?

easbar: Update2: I think I found a better solution here: 933e958d46257890b5c97f375bab797b4a3131cc and now I think the PR is ready to merge.

easbar:
Btw not sure if we already need this, but the Mapbox documentation also describes how to keep the `lat/lng/zoom` state of the map in sync with the application state:

The code is here: https://github.com/mapbox/mapbox-react-examples/tree/master/basic

karussell:
Thanks for investigating this, this works for me. I think this also supersedes #71?

And is it easy to reduce the world bbox to avoid seeing some continents twice?

easbar:
> I think this also supersedes #71?

Yes, I think so.

> And is it easy to reduce the world bbox to avoid seeing some continents twice?

I think we can do this in `Mapbox.ts`:

```javascript
 this.map = new Map({
            // ...
            maxBounds: [[-180, -90], [180, 90]],
            renderWorldCopies: false
        })
```
See also here:
https://docs.mapbox.com/mapbox-gl-js/example/restrict-bounds/
https://docs.mapbox.com/mapbox-gl-js/example/render-world-copies/

> Also this is related to #53.

Not sure. This PR is mostly about how the bbox is applied to the map, but #53 seems also about determining the right bounding box for a route (use appropriate padding, do not zoom in too much etc.)

easbar:
I just disabled the world copies here: b29801b169b0ff797a4ee5ba56e06365589f9150, but feel free to adjust

easbar:
I briefly tried avoiding the initial jump when points are specified (also not fixed in #71), but let's do this in a separate PR. I opened https://github.com/graphhopper/graphhopper-maps/issues/82 for this.

src/App.tsx
Janekdererste: Yes, this is much better. Should we maybe combine those into a single effect though? Since you are using ```route``` in both effects maybe switch to something like
```
useEffect(() => setBbox(route.selectedPath.bbox ? route.selectedPath.bbox : info.bbox), [route, info])
```
I guess this is only a matter of taste though

src/map/layer-group-solid.svg
```diff
@@ -1 +1,5 @@
-<svg aria-hidden="true" focusable="false" data-prefix="fas" data-icon="layer-group" class="svg-inline--fa fa-layer-group fa-w-16" role="img" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 512 512"><path fill="currentColor" d="M12.41 148.02l232.94 105.67c6.8 3.09 14.49 3.09 21.29 0l232.94-105.67c16.55-7.51 16.55-32.52 0-40.03L266.65 2.31a25.607 25.607 0 0 0-21.29 0L12.41 107.98c-16.55 7.51-16.55 32.53 0 40.04zm487.18 88.28l-58.09-26.33-161.64 73.27c-7.56 3.43-15.59 5.17-23.86 5.17s-16.29-1.74-23.86-5.17L70.51 209.97l-58.1 26.33c-16.55 7.5-16.55 32.5 0 40l232.94 105.59c6.8 3.08 14.49 3.08 21.29 0L499.59 276.3c16.55-7.5 16.55-32.5 0-40zm0 127.8l-57.87-26.23-161.86 73.37c-7.56 3.43-15.59 5.17-23.86 5.17s-16.29-1.74-23.86-5.17L70.29 337.87 12.41 364.1c-16.55 7.5-16.55 32.5 0 40l232.94 105.59c6.8 3.08 14.49 3.08 21.29 0L499.59 404.1c16.55-7.5 16.55-32.5 0-40z"></path></svg>
\ No newline at end of file
+<svg aria-hidden="true" focusable="false" data-prefix="fas" data-icon="layer-group"
```
Janekdererste, line 1: Was this change on purpose?

src/map/Map.tsx
```diff
@@ -19,22 +19,42 @@ type MapProps = {
 export default function({ selectedPath, paths, queryPoints, bbox, mapStyle }: MapProps) {
     const mapContainerRef: React.RefObject<HTMLDivElement> = useRef(null)
     const [map, setMap] = useState<Mapbox | null>(null)
+    const prevViewPort = useRef<ViewPort | null>(null)
```
Janekdererste, line 22: I wonder if the functionality to remeber the viewport when the map instance is replaced can be achieved with a functional state update https://reactjs.org/docs/hooks-reference.html#functional-updates 

Here we would get the previous ```map``` instance as parameter to ```useState``` for the map something like this maybe?
```
const mapWrapper = new Mapbox(
            mapContainerRef.current!,
            mapStyle,
            () => {
                setMap(prevWrapper => {
                    mapWrapper.setViewPort(prevWrapper.getViewPort())
                    return mapWrapper
                })
                Dispatcher.dispatch(new MapIsLoaded())
            }
        )
```
On the other hand this might be to late and cause fickering of the map again, because the viewport would be set after the map is loaded.

Janekdererste: Ah, the code example is for line 35, but it also concerns the useRef

Janekdererste (approved):
This looks good overall.

src/App.tsx
easbar: Ah yes this looks short and clean. But are you sure it does the same? Considering there were quite a few edge-cases (for which we have no tests currently :() and I was sure this version works and we might change this again in #87 I think I would rather leave it as is for this PR. Not the best reasons, I know...

src/map/layer-group-solid.svg
easbar: Ah no sorry, this was just auto-formatting, I reverted this change.

src/map/Map.tsx
easbar: Ok interesting. I followed the pattern to use the previous state here: https://reactjs.org/docs/hooks-faq.html#how-to-get-the-previous-props-or-state. If you don't mind also here I would leave as is for now and either change this in a follow-up commit/PR or maybe we do not need to replace the map instance anyway using the react wrapper #87.

Janekdererste: Sorry, for the late answer. I totally agree especially, since we want to replace the map layer anyway.

PR #84, opened 2021-05-20
no error for empty profile

karussell:
The problem is that an empty `profile` parameter is sent to the API endpoint for an URL like:

https://graphhopper.com/maps2/?point=51.47212%2C5.970732&point=49.001769%2C15.804879

When using the `vehicle` parameter this was not a problem as in those cases we set `car` as default. But for profile we require a value


easbar:
Some test is failing now. And maybe use `let` instead of `var`, it does not have some of the strange semantics of `var` and still lets you change the value after initialization (compared to `const`).

karussell:
Ah, sure. I'm not yet used to these JS peculiarities :) ... will also fix the failing tests.

PR #86, opened 2021-05-21
make it possible to translate strings aka i18n

### LINKED ISSUE #41: i18n
introduce a concept where it is easy to internationalize the user facing strings

Reuse existing translations (`web.*`):
https://docs.google.com/spreadsheets/d/10HKSFmxGVEIO92loVQetVmjXT0qpf3EA2jxuQSSYTdU/edit?pli=1#gid=0

@karussell create one JSON file with all translations: https://github.com/graphhopper/graphhopper/tree/master/core/src/main/resources/com/graphhopper/util

Additionally send correct locale parameter to API to get localized instructions.

karussell:
fixes #41


src/translation/Translation.ts
```diff
@@ -0,0 +1,38 @@
...
+        if (!str) return key
+        if (parameters)
+            for (let i = 0; i < parameters.length; i++) {
+                str = str.replace('%' + (i + 1) + '$s', parameters[i])
+            }
+        return str || ''
+    }
+}
+
+let translation: Translation
```
karussell, line 21: singleton similar to how the stores are handled

update-translations.sh
```diff
@@ -0,0 +1,31 @@
...
+  if [[ "$tr" == "SKIP" ]]; then
+    continue
+  fi
+  echo "\"$tr\":{" >> $destination
+  while IFS=$'\r' read -r line; do
+    VAL=$(echo "$line" | cut -f1,$INDEX --output-delimiter="\":\"")
+    if [[ $VAL = web.* ]]; then
+      VAL=${VAL#"web."}
+      echo "\"$VAL\"," >> $destination
+    fi
```
karussell, line 25: here we skip all translation keys without the `web.` prefix - should we do this or import everything?

karussell:
umpf I invested so many minutes to convert the tsv into a json but all I needed is to use the already fetched json from /i18n ... will rewrite.

karussell:
Fetching the i18n via endpoint is nice but the async nature especially when initializing is very ugly. But the biggest problem is that it unnecessarily couples the UI to the server (server side string updates take much longer for a new UI components). I.e. I will revert to the local tr.json file.

karussell:
I just found out about the [Intl](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl) browser API and this project https://projectfluent.org/

Janekdererste:
> Fetching the i18n via endpoint is nice but the async nature especially when initializing is very ugly. But the biggest problem is that it unnecessarily couples the UI to the server (server side string updates take much longer for a new UI components). I.e. I will revert to the local tr.json file. 

I agree that this should be served together with the main bundle. I can't stand websites where the text is loaded separately. And, since the app is already quite heavy weight and the map takes long to load, I think we should try to keep a speedy impression on other things at least

src/translation/Translation.ts
```diff
@@ -0,0 +1,45 @@
+import trJson from './tr.json'
+
+export class Translation {
+    data: any
```
Janekdererste, line 4: This should be of type ``` data: { [index: string] : string ```

src/translation/Translation.ts
```diff
@@ -0,0 +1,45 @@
...
+    fallback: any
```
Janekdererste, line 5: This should be the same type as data

src/translation/Translation.ts
Janekdererste: The other stores are not real singletons though. I didn't make the singletons because they are a little easier to unit test if dependency injection still works

src/translation/Translation.ts
```diff
@@ -0,0 +1,45 @@
...
+            }
+        return str || ''
+    }
+}
+
+let translation: Translation
+export function setTranslation(lang: string, overwrite = false): Translation {
+    if (translation && !overwrite) throw new Error('translation already initialized')
+
+    let json = (trJson as any) as Record<string, any>
```
Janekdererste, line 32: I would think that you only need to cast once ``` const json = trJson as Record<string, any>```

Janekdererste: I guess any as record value is fine here, because we don't really care in this scope

src/translation/Translation.ts
```diff
@@ -0,0 +1,45 @@
...
+    fallback: any
+
+    constructor(data: any, fallback: any) {
+        if (!data) throw Error('Translation is undefined')
+        if (!fallback) throw Error('Translation fallback is undefined')
+        this.data = data
+        this.fallback = fallback
+    }
+
+    tr(key: string, parameters?: string[]): string {
```
Janekdererste, line 14: maybe call this ```get```? Then one could call this like ```translation.get("bla")```

I find this easier to pronounce when reading it instead of ```translation.tr("bla")```

Janekdererste (approved):
I don't really see, what the status of this is, but it looks fine. Maybe change some translation in the actual app's code, to see wehther it works well like this?

karussell:
> The other stores are not real singletons though.

Can you explain what you mean is the difference to Stores where a similar pattern is used

`"let mapOptionsStore: MapOptionsStore"`

> I didn't make the singletons because they are a little easier to unit test if dependency injection still works

:+1:

src/translation/Translation.ts
karussell: Ok. Just fyi: the idea would be to make the translation object rather short like `t` so that it gets `t.get` as we will have a translation probably in many places.

Janekdererste:
> Can you explain what you mean? In Stores a similar pattern is used

The store modules export their class. Then in the Stores module they are instantiated and stored in a global variable. So, they are not Singletons in the sense that one can still instantiate those classes from outside. In the application they are used as if they were Singletons. Again, did this for easier testing. 

I think the Translation is different though and can be a singleton.

karussell:
And you mean we should prevent instantiation of Translation from outside? (because it can be a singleton)

src/translation/Translation.ts
karussell: Have changed it to be more precise.

karussell:
> Maybe change some translation in the actual app's code, to see wehther it works well like this?

Have used it for the profile names and input hints.

update-translations.sh
karussell: Let's keep it separated

Janekdererste:
> And you mean we should prevent instantiation of Translation from outside? (because it can be a singleton)

I don't have an oppinion on that. I just meant that the translation is somewhat self contained. Since it will take the language parameter from somewhere and build a translation instance based on that, I don't really see a need for outside access. You may go ahead as you like :-)

karussell:
Ok, will merge then. Also feel free to change this aspect in the future :)

src/translation/tr.json
```diff
@@ -0,0 +1,1760 @@
+{"en_US":{
```
easbar, line 1: I think this should be `"en-US"` instead? At least it does not work on my laptop this way.

src/translation/Translation.ts
```diff
@@ -0,0 +1,45 @@
...
+}
+
+let translation: Translation
+export function setTranslation(lang: string, overwrite = false): Translation {
+    if (translation && !overwrite) throw new Error('translation already initialized')
+
+    let json = trJson as Record<string, any>
+    for (let property in json) {
+        if (property.startsWith(lang)) {
+            translation = new Translation(json[property], json['en_US'])
```
easbar, line 35: Also here `'en-US'`?

easbar: Hm, but many other languages also use underscore?!

karussell: I think we need some matching like we have somewhere in old gh maps: https://github.com/graphhopper/graphhopper-maps/issues/97

Issue #81, opened 2021-05-20
Replace Mapbox GL JS with Maplibre GL JS


karussell:
code: https://github.com/MapLibre/maplibre-gl-js
docs: https://maplibre.org/maplibre-gl-js-docs/api/

especially when we hit a bug or want to be mentioned there :)

HarelM:
+1
If you have any questions feel free to ping me. I'm one of the core contributers at maplibre :-)

easbar:
We use OpenLayers since #159

HarelM:
Cool!
Bare in mind that when I tested OL about two years ago the performance for vector tiles on mobile phones was very poor.
But it might not be a strong argument for you guys.
Also note that we are planning on adding terrain 3d support in maplibre in the near future which I don't think is possible in OL.
But I know OL is a solid choice :-) If I wasn't interested in vector tiles I would surely use them.
Thanks for the update.

easbar:
Yes this was also a major decision point for us. For now we are happy to continue using raster tiles as we did in the past.

Issue #80, opened 2021-05-20
Problems with map style selection on firefox mobile


Janekdererste:
selecting a different map on mobile nearly always results in selecting kurviger first before I can open it

Janekdererste:
@karussell I think you reported this in #74 . I can't reproduce this. I this still an issue for you? It seems like the radio button selection is not working properly?

karussell:
Indeed this works now. Strange.

easbar:
Maybe this was fixed by #77 (?)

Issue #78, opened 2021-05-20
Improve layer management


easbar:
Currently we maintain a hard-coded list of layers in `Mapbox.ts`: https://github.com/graphhopper/graphhopper-maps/blob/d8603138930fca949adb1c94264e6479162852fb/src/map/Mapbox.ts#L17-L20

This seems to work for now, but when we start adding more layers this is probably not ideal and it might be better to keep `Mapbox.ts` independent from the layers we like to add to the map. With some kind of external layer management maybe we also do not need the current workaround of re-creating the map instance when we change the background layer (mapStyle) here: 

https://github.com/graphhopper/graphhopper-maps/blob/d8603138930fca949adb1c94264e6479162852fb/src/map/Map.tsx#L22-L35

Maybe we can take a look at popular react/mapbox wrappers to borrow some ideas for this and/or even use one of those wrappers, like:

* https://github.com/visgl/react-map-gl
* https://github.com/alex3165/react-mapbox-gl

It's probably a good idea to review this issue when we start working on #24

easbar:
With #87 it is now easier to add more layers and currently I don't think there is anything more to be done.

PR #87, opened 2021-05-25
Mapbox/React wrapper

### LINKED ISSUE #82: Do not show flicker/jump when map loads with route coordinates
For example when we open the app with this url: `http://localhost:3000/?point=48.098438%2C15.203637&point=47.755246%2C17.388894&profile=car` we first see the map and then the map jumps to the route.

See also discussions in #53, #71, #77.

easbar:
I was curious and tried Uber's Mapbox/React wrapper: https://visgl.github.io/react-map-gl/

~~This is just experimental and not meant to be merged in the current state. This branch is also based on top of #77 and #75, but that should not be so important. I just did not want to do the migration without the fix in #77 and before the path detail code was moved out of the map component in #75.~~ 

The philosophy of the wrapper is interesting: https://visgl.github.io/react-map-gl/docs#design-philosophy

The wrapper is meant to address the problem that Mapbox is stateful and therefore does not integrate so well with React. We already faced some of the related issues like switching the background layer without re-creating the Map component or changing the bounding box in #72 etc. So I was hoping using this wrapper the viewport updates and the style switching feels more natural. 
The migration actually went relatively smooth and I did not run into bigger issues yet. The state handling definitely feels better to me and it seems like less workarounds are required.

Other advantages could be that according to the documentation the wrapper works with mapbox-gl 1&2 and maplibre as well (I only tried mapbox-gl 1 so far) (#81)  and it will be easier to move the different map layers out of the central map component (#78), just like I already moved out the viewport state handling. Another advantage could be easier integration with deck.gl (but I am just speculating and not sure if we even need this). Anyway, so far I do not see many reasons *not* to use the wrapper. At least we should keep this in mind and maybe comeback here when we see more issues related to the Mapbox state management for example.

I will update this branch when #75 and #77 are finished so it will be easier to track the changes. Otherwise, there are the following todos on this branch to make it equal to master/#75/#77, currently (feel free to edit this in case you notice something):

* [x] ~~use actual marker symbols for query points and the path detail hovering. this should not be a real problem, but I just could not find any default marker svgs to use. For the path detail marker we might need to adjust the code because I could not find a way how to use the HTML node we get from heightgraph currently as marker.~~ improve look of the point markers and the path detail marker. so far I use a rather ugly point marker and just open a popup for the path detail marker. -> postponed to #106 
* [x] put back the long touch handler (I just commented it out for now)
* [ ] tune the fly to animation (it looks like its not entirely the same as before yet)
* [x] the Popup is missing the closeOnMove option
* [x] missing maxBounds option for the map (it zooms out too much)
* [x] should we move each map layer to a separate file/component so `Map.tsx` does not have to deal with all the different things? I think so and did this in #108 

On the other hand this branch already fixes #82 (and maybe #80 as well) and it also implements the missing 'center map' context menu entry.

Btw there is also another such wrapper here: https://github.com/urbica/react-map-gl


src/map/Map.tsx
```diff
@@ -1,41 +1,290 @@
...
+            <PopupComponent
+                coordinate={popupCoordinate}
+                queryPoints={queryPoints}
+                onSelect={() => setPopupCoordinate(null)} />
+        </Popup>}
+        {createQueryPointMarkers(queryPoints)}
+        {pathDetailPoint && createPathDetailMarker(pathDetailPoint)}
+        {createUnselectedPaths(currentPaths)}
+        {createSelectedPath(selectedPath)}
+        {createHighlightedPathSegments(highlightedPathDetailSegments)}
```
easbar, line 98: In principle we could replace this with something like `{...layers}` and the `layers` would be passed into the map component as prop. This way we could just add different layers to the map without modifying the `Map` component. Then again, I wonder what there would be really left for the `Map` component to do. The `react-map-gl` kind of *is* our map component already... One important question might be what should be re-used in case for example we wanted to 'switch' the app to isochrones or map-matching.

Janekdererste:
This looks quite nice I think. If it just works, maybe go ahead and replace our Mapbox class with the component?

easbar:
I basically replaced our Map component and our Mapbox class with the new Map component that uses the wrapper. With the wrapper it seems like we do not need both `Map.tsx` and `Mapbox.ts`. But anyway we should get #75 and #77 merged before and if you want I can then move some of the new Map code into Mapbox.ts so the diff becomes more readable. Currently this PR does not just work. See the todo list in my initial comment.

easbar:
I think this PR is already in a good state. We could improve the look of the markers also after merging it IMO and do this together with #106. Should we move the different map layers to separate components already (related #78). They already are components so all we need to do is move them to separate files. For example we could create something like:

* ContextMenuLayer 
* QueryPointsLayer 
* PathsLayer
* PathDetailLayer (marker + highlighted path segments)

> missing maxBounds option for the map (it zooms out too much)

I'd like to get this done for this PR ideally, but unfortunately I could not make this work yet. I think we have to restrict the viewport in `ReactMapGL#onViewportChange`, but the tricky thing is to obtain the zoom level for the given width and height so the maximum browser window space is used, but we do not show 'dead space' left and right of the map as we currently do in this branch:

image

> tune the fly to animation (it looks like its not entirely the same as before yet)

I don't have a strong preference here, but feel free to adjust this. This should be possible using the options in `ViewportStore#reduce(SetViewportToBbox)`.

easbar:
> so all we need to do is move them to separate files

Haha, except this seems more complicated than I thought. I still find this very appealing, but it seems like we should do this in a follow-up PR for #78.

easbar:
> I think we have to restrict the viewport in ReactMapGL#onViewportChange

This is working now. I think this PR is ready to go

Janekdererste:
When I load the map without parameters, the bounding/viewport is a little off 
grafik

Janekdererste:
When I load on full screen i get the following error:
```
Uncaught Error: `longitude` must be supplied
    at assert (assert.js:6)
    at new MapState (map-state.js:74)
```

easbar:
> When I load on full screen i get the following error:

I had the same error on initial map load here: a9ab9a792ee7d19d602a7d4057eef7fbb19634d1, but only in Chrome and not in Firefox. How do you load full screen? Do you mean browser full screen or Mapbox full screen (not implemented yet I thought, #20).

easbar:
> When I load the map without parameters, the bounding/viewport is a little off

Are you sure you use the latest commit? I fixed this in one of my last commits.

src/App.tsx
```diff
@@ -54,48 +64,55 @@ export default function App() {
...
 
     const isSmallScreen = useMediaQuery({ query: '(max-width: 44rem)' })
-
-    const [bbox, setBbox] = useState<Bbox>([-180, -90, 180, 90])
     // todo: maybe combine these effects into one? see discussion in #77
     useEffect(() => {
         // make sure the path bbox and the path details bbox take precedence over the info bbox
-        if (!route.selectedPath.bbox && !pathDetails.pathDetailBbox) setBbox(info.bbox)
+        if (!route.selectedPath.bbox && !pathDetails.pathDetailBbox && info.bbox.every(num => num !== 0))
+            Dispatcher.dispatch(new SetViewportToBbox(info.bbox))
```
Janekdererste, line 76: Dispatches should only be caused by external events, such as User Interaction or callbacks issued by the browser. 

This effect is also going to be executed when a store issues a change event caused by another dispatch. This is not supposed to happen. The original Dispatcher implementation of flux forbids this. I guess it is time that we add this to our Dispatcher as well. 

Instead the Viewport Store should listen to the original Actions which may cause changes of the viewport such as RouteReceived or InfoReceived. The store should then decide what the bbox should be and set this in its state. This follows the pattern, that the stores manage the state and the components decide how to represent this state. This pattern is violated here.

src/map/Map.tsx
```diff
@@ -1,65 +1,130 @@
...
-}: MapProps) {
-    const mapContainerRef: React.RefObject<HTMLDivElement> = useRef(null)
-    const [map, setMap] = useState<Mapbox | null>(null)
-    const prevViewPort = useRef<ViewPort | null>(null)
-    useEffect(() => {
-        prevViewPort.current = null
-    }, [bbox])
+export default function ({ viewport, mapStyle, queryPoints, mapLayers }: MapProps) {
+    const [popupCoordinate, setPopupCoordinate] = useState<Coordinate | null>(null)
+    const longTouchHandler = new LongTouchHandler(e => setPopupCoordinate({ lng: e.lngLat[0], lat: e.lngLat[1] }))
```
Janekdererste, line 21: I guess this could be modeled with an effect now. I chose to model this as a class because this Handler lived outside of a react component. We can also leave it like this for now. 

I'm not sure though what happens if a touch down event happens and the component gets re-rendered, before the timout finishes. The current handler implementation doesn't have a cancel function

src/map/Popup.tsx
```diff
@@ -1,60 +1,20 @@
...
-        this.popup.setLngLat(coordinate).setDOMContent(this.container).addTo(this.map)
-    }
-
-    hide() {
-        this.popup.remove()
-    }
-}
-
-function PopupComponent({
+export function PopupComponent({
```
Janekdererste, line 8: Ah, this becomes much easier!

Janekdererste (changes requested):
Looks very nice overall. I think this is defenitely the right way to go.

The dispatches from App.tsx must be changed though.

Janekdererste:
> Are you sure you use the latest commit? I fixed this in one of my last commits. This one specifically: 7623366

I din't

Janekdererste:
> How do you load full screen?

I mean the window being sized to full screen, but the frame still showing

easbar:
> I mean the window being sized to full screen, but the frame still showing. It happened on both, Firefox and Chrome

I cannot reproduce this? Do you still see this since you updated?

src/App.tsx
easbar: Ah ok thanks for your explanation, I need to learn more about this then. Should I try to improve this or do you rather do it yourself (and I will be watching you of course :))

src/map/Map.tsx
easbar: Maybe we can continue here in #111.

Janekdererste:
> I cannot reproduce this? Do you still see this since you updated?

It seems to be gone now :-)

easbar:
> It seems to be gone now :-)

You probably pulled the broken version, this problem existed for a few commits and I only fixed it today, because it did not show up in Firefox for some reason.

easbar:
Btw the bounding box is now more fixed than in master. In master we see empty space above the map when we scroll vertically (and zoom out too much)

Janekdererste:
> Btw the bounding box is now more fixed than in master. In master we see empty space above the map when we scroll vertically (and zoom out too much)

I like this very much. The white bars were terrible.

src/App.tsx
Janekdererste: I can try to do this.

src/map/Map.tsx
Janekdererste: Ah, yes. I didn't notice this issue yet

karussell:
This is nice :+1:

src/App.tsx
easbar: Ok, thanks.

Janekdererste: So the order is:
- 1. pathDetails bbox
- 2. selectedpath bbox
- 3. info bbox
- 4. default bbox

?

easbar: Yes. But default bbox is just 'world' I think. I have been using this little checklist:

```
app start -> world
/info -> map bbox
/route -> path bbox
clear route -> keep current bbox
change layer -> keep current bbox
select path detail range -> path detail bbox
clear path detail range (click diagram) -> path bbox
```

And I also noticed another bug in master: When we select a path detail range the bbox zooms in, but when we click another alternative in the instructions list then the map does not fly to the other path, probably because the selected path bbox does not update. Not crazy important, but if you change the code here maybe this problem will be solved as well.

Janekdererste:
So, I changed the Dispatch stuff and moved everything into the Viewport store. 

I also noticed that there is no easing on zoom. Do we need to specify an easing function? The examples on the mapbox component page seem to have a smooth animation when I zoom by mouse wheel.

Janekdererste:
Unit testing? At least of the store?

easbar:
> I also noticed that there is no easing on zoom. Do we need to specify an easing function? The examples on the mapbox component page seem to have a smooth animation when I zoom by mouse wheel.

Yes, see the end of this comment https://github.com/graphhopper/graphhopper-maps/pull/87#issuecomment-861418655 and the todo in the `calculateLatLngFromBbox` function. I just wasn't sure what we want here, but feel free to tune this.

easbar:
> So, I changed the Dispatch stuff and moved everything into the Viewport store.

Thanks for this. To me it seems a bit strange that the ViewportStore knows about all the routesuccess, pathdetail etc. actions. Changing the viewport rather seems like a basic feature of the map that is used by different more specific features (like routes and path details). So I would have expected the dependency the other way around: The route store (same with path detail) is a client of the viewport store. For example when we add more features that modify the viewport I would rather add more client code that uses the viewport store than modify the viewport store's reduce function every time.

Janekdererste:
> To me it seems a bit strange that the ViewportStore knows about all the routesuccess, pathdetail etc. actions

That's an intersting question. I've thought of a store as something which holds and manages a portion of the app's state. The store reacts to Actions happening in the app and updates its state accordingly. So, when we want the viewport to be different after a certain event occured, the ViewportStore should take care of it. Also, every action is dispatched to every store anyway, so that every store knows about every action already.

> I would rather add more client code that uses the viewport store than modify the viewport store's reduce function every time.

I think the set of actions our app has should represent the events that happen in our app. Introducing actions like the SetViewportFromBbox action don't actually represent an action of the app but a desired change to the state caused by something else ( a successfull routing request for example)

easbar:
Ok, I understand there is a difference between 'reacting to a state change' and sending an event like `SetViewportFromBBox`. So feel free to leave it like this for now and I might think about this a bit more.

> Also, every action is dispatched to every store anyway, so that every store knows about every action already.

No, I don't think so. Every action is passed to every store yes, but the code of the different stores is not coupled to all actions and the stores do not contain any code related to actions they do not handle explicitly. We can add/remove actions without changing all the stores. But now we cannot remove the path detail stuff (for example not needed for isochrone) without adjusting the (IMO unrelated) viewport changing code.

src/App.tsx
easbar: > And I also noticed another bug in master: When we select a path detail range the bbox zooms in, but when we click another alternative in the instructions list then the map does not fly to the other path, 

Nice, this is also fixed now :+1:

easbar:
> Unit testing? At least of the store?

I added a test in my last commit. It is actually quite nice that we can test the viewport transitions on this level now.

jest.config.js
```diff
@@ -1,5 +1,5 @@
 module.exports = {
     preset: 'ts-jest',
-    testEnvironment: 'node',
+    testEnvironment: 'jsdom',
```
easbar, line 3: Is this ok? I needed this for the viewportstore test as it uses `window.matchMedia`.

test/stores/ViewportStore.test.ts
```diff
@@ -0,0 +1,51 @@
...
+                )
+            )
+            expect(state.longitude).toBeCloseTo(10.5, 1)
+            expect(state.latitude).toBeCloseTo(50.9, 1)
+            expect(state.zoom).toBeCloseTo(8.8, 1)
+            // ... select path details
+            state = store.reduce(state, new PathDetailsRangeSelected([10.46, 50.88, 10.47, 50.9]))
+            expect(state.longitude).toBeCloseTo(10.46, 2)
+            expect(state.latitude).toBeCloseTo(50.89, 2)
+            expect(state.zoom).toBeCloseTo(13.7, 1)
```
easbar, line 43: I really like we can do this now.

test/matchMedia.mock.ts
```diff
@@ -0,0 +1,11 @@
+Object.defineProperty(window, 'matchMedia', {
+    writable: true,
+    value: jest.fn().mockImplementation(query => ({
```
easbar, line 3: I needed this because viewport store uses `window.matchMedia`.

src/stores/ViewportStore.ts
```diff
@@ -0,0 +1,131 @@
...
+    minPitch?: number
+    maxPitch?: number
+    transitionDuration?: number
+    transitionEasing?: Function
+    transitionInterpolator?: TransitionInterpolator
+    transitionInterruption?: number
+}
+
+// have this right here for now. Not sure if this needs to be abstracted somewhere else
+const mediaQuery = window.matchMedia('(max-width: 640px)')
```
easbar, line 35: This makes testing the store a bit complicated, see my other comments. Can we just inject this into the constructor? And is it the same as `isSmallScreen` in `App.tsx`? Here we use `640px` and in `App.tsx` we have something similar with `44rem`.

src/stores/ViewportStore.ts
```diff
@@ -49,48 +62,64 @@ export default class ViewportStore extends Store<ViewportStoreState> {
...
               right: 16,
               left: 16,
           }
         : {
               top: 100,
-              bottom: 100,
+              bottom: 300,
               right: 100,
               left: 500,
           }
```
easbar: Nice! This fixed the problem that sometimes the markers were hidden by the path detail diagram, see #53

karussell:
This seems to solve several issue like @easbar said also the issue where the elevation widget hides the routes. It zooms also to the route when you select the alternative route which is really nice.

There is a minor problem with the path details widget (not too important :) ) but it is hard to describe. Sometimes the elevation/pathdetails data does not show up when the mouse goes over the area (try left and right). There seem to be areas that just do not work.

Hm ... the path detail widget with current mapbox seems to have this problem too but only for small areas of the widget.

easbar:
> also the issue where the elevation widget hides the routes

Yes, even though this was unrelated to the wrapper: https://github.com/graphhopper/graphhopper-maps/pull/87#discussion_r654781011

src/stores/ViewportStore.ts
Janekdererste: Yes, it would be better to have this media query at 44rem as well. And yes, it is supposed to be the same as ```isSmallScreen```.

I guess you could even inject something like ```isSmallScreen: () => boolean``` and pass ```new ViewportStore(() => window.matchMedia(...))``` in ```index.tsx```. Then in the test you could supply your own predicate.

test/matchMedia.mock.ts
Janekdererste: See comment below.

test/stores/ViewportStore.test.ts
Janekdererste: Yes, this looks very nice.

jest.config.js
Janekdererste: I have no preference here. I think I just put in, what the "getting-started-with-jest" tutorial says.

easbar: I reverted. jsdom sounds more correct but who knows and as long as it works with the current value no need to change

src/stores/ViewportStore.ts
easbar: I did this. The only thing is that we have this media query twice now. One time in `index.tsx` where we use `window.matchMedia` and one time in `App.tsx` where we use the `useMediaQuery` hook. We cannot use the hook in `index.tsx` because it would be outside of any react component and I thought there is probably a reason you use the `useMediaQuery` hook in `App.tsx`

karussell:
I just played chaos monkey a bit more and now an error occurred when using the same coordinates (like #36). But now something like "e is undefined". To prevent this from re-introducing it again, would we need something like playwright or cypress or is this possible in unit tests?

easbar:
This error occurs because we do this on the server-side:
java
```
 if (p.getPoints().size() >= 2) {
     jsonPath.putPOJO("bbox", p.calcBBox2D());
 }
```                
so when the points are identical the bbox field is missing and we do not handle this yet. Things like this can be tested without UI/browser tests.

easbar:
Should we change this on the server-side? It seems a bit ugly that the bbox is missing in this case. Why not just return [lat, lon, lat, lon] in case we request a route from (lat, lon) to (lat, lon)?

easbar:
I made the error more visible by using a debug build for aws: d731b67d9add4c328bfe2f2d06d3a86f6a21b4aa and logging errors for /route requests: fcfccc5860de50f92a653c18d8147e7af7b85e63

src/stores/ViewportStore.ts
Janekdererste: Yes, the hook is used, so that the layout changes when the window is resized. 

There are several other places in css where this media query is used as well. I think as long as we only have one breakpoint we may be fine.

karussell:
I wasn't able to find the reason but will change this on the server side as this made problems before. The oldest location seems to be [here](https://github.com/boldtrn/graphhopper/blob/5935f5bd94969d236ae263e1c20453098373394b/web/src/main/java/com/graphhopper/http/SimpleRouteSerializer.java).

Issue #70, opened 2021-04-25
server side timeout of alternative routes results in JS error


karussell:
error in the UI is

```
(intermediate value).hints is undefined
```

but it should better show "500 Internal Server Error" or something.

Example: https://graphhopper.com/maps2/?point=51.11452800111718%2C10.591936151781283&point=48.8588897%2C2.320041027461107&profile=foot (probably alternative route calculation takes too long for this distance.)

Janekdererste:
I guess this could become a more general issue about error messages. So far errors are caught and an Error Action is dispatched. Then we have an Error Store, which just swallows the dispatched error message. This could be more elaborate I guess.

PR #88, opened 2021-05-26
#32 autocomplete keyboard

### LINKED ISSUE #32: Use up/down cursor for autocomplete selection

### LINKED ISSUE #54: leaving input with tab should work (keyboard navigation)
When I type a location in the first input, then "tab" into second, then type a location in the second input and hit return it should show the route. Currently it is required that I select a location from both suggestion lists to update the route.

Also when copy and pasting a location as coordinate it does not route although no geocoding needs to be done.

### LINKED ISSUE #69: introduce a small 100ms delay for geocoding
Currently every changed character results in a geocoding request. E.g. if I delete a location with 10 characters with backspace it will trigger all 9 requests. If there would be a delay then this would be only 1 or 2 reducing unnecessary load (especially if deployed to many :) )

Janekdererste:
fixes #32 
fixes #54 
fixes #69 

Overhauled the Address Input Component. 

- It should now be possible to use arrow keys to navigate the result list. 
- On Enter the currently selected element from the result list is selected or the first one if no arrow key was pressed before
- On small screens the fullscreen view can be left with Esc
- When the Text Input looses focus, the geocoding results are not shown anymore
- Added a debounce of 500ms. (Maybe that's too much?)

### DIFF
--- a/src/sidebar/search/AddressInput.tsx
+++ b/src/sidebar/search/AddressInput.tsx
@@ -117,6 +117,7 @@ function calculateHighlightedIndex(length: number, currentIndex: number, increme
  */
 class Geocoder {
     private requestId = 0
+    private timeout = new Timout(500)
     private api = new ApiImpl()
     private onSuccess: (hits: GeocodingHit[]) => void
 
@@ -125,19 +126,22 @@ class Geocoder {
     }
 
     request(query: string) {
-        const currentId = this.getNextId()
+        this.requestAsync(query).then(() => {})
+    }
 
+    async requestAsync(query: string) {
+        const currentId = this.getNextId()
+        this.timeout.cancel()
         if (!query) return
 
-        this.api
-            .geocode(query)
-            .then(result => {
-                const hits = Geocoder.filterDuplicates(result.hits)
-                if (currentId === this.requestId) this.onSuccess(hits)
-            })
-            .catch(reason => {
-                throw Error('Could not get geocoding results because: ' + reason)
-            })
+        await this.timeout.wait()
+        try {
+            const result = await this.api.geocode(query)
+            const hits = Geocoder.filterDuplicates(result.hits)
+            if (currentId === this.requestId) this.onSuccess(hits)
+        } catch (reason) {
+            throw Error('Could not get geocoding results because: ' + reason)
+        }
     }
 
     private getNextId() {
@@ -154,5 +158,23 @@ class Geocoder {
             }
             return false
         })
+    }
+}
+
+class Timout {
+    private delay: number
+    private handle: number = 0
+
+    constructor(delay: number) {
+        this.delay = delay
+    }
+
+    wait() {
+        return new Promise(resolve => {
+            this.handle = window.setTimeout(resolve, this.delay)
+        })
+    }
+    cancel() {
+        clearTimeout(this.handle)
     }
 }
--- a/src/sidebar/search/AddressInput.tsx
+++ b/src/sidebar/search/AddressInput.tsx
@@ -77,7 +77,6 @@ export default function AddressInput(props: AddressInputProps) {
                     onKeyDown={onKeypress}
                     onFocus={() => setFullscreen(true)}
                     onBlur={() => {
-                        console.log('on blur')
                         setFullscreen(false)
                         setGeocodingResults([])
                     }}
--- a/src/sidebar/search/AddressInput.tsx
+++ b/src/sidebar/search/AddressInput.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useRef, useState } from 'react'
+import React, { useCallback, useEffect, useRef, useState } from 'react'
 import { QueryPoint } from '@/stores/QueryStore'
 import { GeocodingHit } from '@/api/graphhopper'
 import GeocodingResult from '@/sidebar/search/GeocodingResult'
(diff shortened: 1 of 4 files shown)


src/sidebar/search/AddressInput.tsx
```diff
@@ -15,43 +15,75 @@ export interface AddressInputProps {
...
+    const [fullscreen, setFullscreen] = useState(false)
     const containerClass = fullscreen ? styles.container + ' ' + styles.fullscreen : styles.container
+
     return (
         <div className={containerClass}>
             <div className={styles.inputContainer}>
                 <input
                     className={styles.input}
-                    type='text'
+                    type="text"
```
easbar, line 70: maybe the code formatting settings were off here?

src/sidebar/search/AddressInput.tsx
```diff
@@ -84,27 +125,30 @@ class Geocoder {
...
+        }
     }
 
     private getNextId() {
         this.requestId++
         return this.requestId
     }
 
-    private static filterDuplicates = function(hits: GeocodingHit[]) {
+    private static filterDuplicates = function (hits: GeocodingHit[]) {
```
easbar, line 151: ... same here

easbar (commented):
Nice, this is very handy :+1:

Can we allow entering GPS coordinates directly as mentioned in #54? And should we (not) also trigger an update when we press Enter again?

Btw I just noticed this (triple?) error message:

image

easbar:
Probably also a bit off-topic, but the profile selection seems to contain some invalid options: 

image

karussell:
> Probably also a bit off-topic, but the profile selection seems to contain some invalid options:

This is also a problem on old GraphHopper Maps and will be fixed once the latest shared-outdoor is deployed (that removes some custom profiles that I wanted to test out)

src/sidebar/search/AddressInput.tsx
Janekdererste: I think in HTML-Tags, the formatting uses ". Everywhere else ' is used.

src/sidebar/search/AddressInput.tsx
Janekdererste: Yes, this is weird.

src/sidebar/search/AddressInput.tsx
easbar: Hm, when I run 'Reformat Code' in IntelliJ it changes the quotes to single quotes

easbar: Ugh sorry, I did not know 'Reformat Code' and 'Reformat with Prettier' are two different commands even after enabling the Prettier code style: https://intellij-support.jetbrains.com/hc/en-us/community/posts/360002745119-How-to-have-Reformat-Code-feature-behave-exactly-like-Prettier-

Using `Reformat with Prettier` the quotes are actually changed to double quotes, unless I add ` "jsxSingleQuote": true` to `prettierrc`.

src/sidebar/search/AddressInput.tsx
easbar: This also seems to be different for IntelliJ 'Reformat Code' and 'Reformat with Prettier'. The first removes the space and the latter adds it...

What is your workflow for this? Normally I sometimes press 'Alt+Shift+L' to reformat the code in case I want it to be formatted, but then I am just using the 'Reformat Code' checkbox in the Git commit Dialog to make sure whatever I am committing is formatted correctly. I could easily remap the shortcut to the 'Reformat with Prettier' action, but unfortunately there is no equivalent before commit action in the commit dialog. So I would have to setup some kind of external pre-commit logic to achieve the same effect. Or maybe tweak the IntelliJ reformat code rules to match the prettier style (but that seems rather fragile).

src/sidebar/search/AddressInput.tsx
Janekdererste: Oh no! 🙄

src/sidebar/search/AddressInput.tsx
Janekdererste: I have set prettier to reformat the code on safe. So basically every file which I'm about to commit is already properly formatted. 

What I'm missing though is automated import optimization. This is something the commit dialog also does for me, but with prettier reformatting on save things get tricky because intellij removes unused imports before the commit, then saves, then prettier kicks in and then I get a warning, that the files I'm about to commit have changed...

The way I have it now, is that I don't do any auto-reformating on commit, but only run code inspection. IntelliJ warns me about unused imports and I remove them manually. It doesn't happen too frequently, but I find it still anoying that I have to take care of imports.

Janekdererste:
> Btw I just noticed this (triple?) error message:

This is because we have three errors coming from the backend. The Error handler just connects each error with an "and". This could be improved in general. Together with unhelpful messages if no route was found. Related to #70

src/sidebar/search/AddressInput.tsx
easbar: Ok thanks, I'll try this as well then. We probably need to re-format the entire code base again then, because when I did this I used 'Reformat code' instead of 'Reformat with Prettier' because I did not know they weren't the same... Maybe this is also why @karussell saw different formatting when using prettier in vscode...

Janekdererste:
> And should we (not) also trigger an update when we press ...

I don't think so. Why trigger anything if nothing has changed?

easbar:
> This is because we have three errors coming from the backend. The Error handler just connects each error with an "and". This could be improved in general. Together with unhelpful messages if no route was found. Related to #70

Ah ok, so let's fix this in #70 then.

> I don't think so. Why trigger anything if nothing has changed?

Hm, my use-case is often debugging (I need to trigger the request to check something on the server-side). So rather special I know :) But then again, as a user I also somehow feel like I want to 'control' the app, so in case I push Enter I expect something to 'happen'... For example when I move the map, pushing Enter should also scroll the map back to the path (?)

Janekdererste:
> Hm, my use-case is often debugging (I need to trigger the request to check something on the server-side). So rather special I know :) But then again, as a user I also somehow feel like I want to 'control' the app, so in case I push Enter I expect something to 'happen'... For example when I move the map, pushing Enter should also scroll the map back to the path (?)

Okay this makes sense then. Doing it from within the address search component is not so trivial though. Which behaviour do we want in this case? Simply try to use the text-input's content as coordinate? Only if it has changed since the last time? Or should the Enter -> Search thing be a more global thing? Also, should the search be triggered only in case an input box is highlighted or again at any time?

While writing this it sounds like this is a global functionality. If it is okay for you, I would put this off until later and put it into another issue. Otherwise we have to discuss first I guess.

easbar:
Sure! This PR already improves so many things, no need to do everything at once.

Issue #57, opened 2021-04-13
Enable commented out map layers


easbar:
See discussion in https://github.com/graphhopper/graphhopper-maps/pull/55

karussell:
Let's close here for now.

PR #89, opened 2021-06-03
uncomment osm.de and added mapilion vector tiles

karussell:
I investigated a bit these CORS issues and for OSM.de it was a problem that their redirection did not contain the CORS headers and failed (combined with the fact that we did not used https). But when using the correct URLs that are not redirected this works now.

The other layers do not have a proper CORS support it seems and probably do not want to be included elsewhere.

The performance of the sorbian layer is very bad (in Firefox/Linux) and so I left it commented out. My suspicion is that it uses some kind of redirect mechanism to fetch normal tiles for the area outside of Germany/Lusatia and fetches Sorbian tiles only for Germany/Lusatia and maybe the mapbox client is not able to handle this redirection trick properly.

I included the mapilion vector tiles as alternative to MapTiler. I also tried the [vector tiles from Thunderforest](https://www.thunderforest.com/docs/vector-maps-api/) but I got JS errors.

Then I replaced neighbourhood thunderforest layer with atlas as it is nicer and includes house numbers.

Two questions - maybe we should clarify them in a different issue:

 - [x] Should we add the `2x` to the URLs for bigger devices instead of hardcode them? Yes, but different issue and for higher dpi not bigger devices.
 - [x] formatting
 - [x] add nice Atlas layer from thunderforest


easbar:
> Unfortunately the formatting kicked in - should I revert this and we do some formatting for the whole project once or automatically somehow for every commit?

Yes, sorry I auto-formatted everything using the wrong formatting rules (see here: https://github.com/graphhopper/graphhopper-maps/pull/88#discussion_r643434996). So let's do the formatting again in all branches?

easbar:
This is related to #57

karussell:
Ok, created #90 for formatting the master branch.

> Yes, sorry I auto-formatted everything using the wrong formatting rules (see here: #88 (comment)). 

Ok, yeah, this is all a bit confusing. Also in VSCode.

Janekdererste:
> Should we add the 2x to the URLs for bigger devices instead of hardcode them?

I always thought this is for high dpi devices regardless of the screen size? I think we could also test for this as well of course.

karussell:
Ah, ok. Yes, sounds more logical :D

PR #90, opened 2021-06-03
format files with npx prettier

karussell:
I just ran

```
npx prettier --write "**/*.ts"
```

And this also for tsx and css.

I tried a few files in VSCode and it seems to be identical. (there seems to be [no "format all files" in VSCode](https://stackoverflow.com/questions/43666270/how-do-i-format-all-files-in-a-visual-studio-code-project))


easbar:
LGTM and no changes compared to doing the same in IntelliJ. Maybe format *.html as well?

src/index.html
```diff
@@ -1,29 +1,29 @@
...
-            height: 100%;
-            width: 100%;
-            margin: 0;
-            padding: 0;
-            font-family: 'Segoe UI', Tahoma, Geneva, Verdana, sans-serif;
-        }
+<html lang="en">
+    <head>
+        <meta charset="UTF-8" />
+        <meta content="width=device-width, initial-scale=1" name="viewport" />
```
karussell, line 5: Interesting. Here it uses " instead of ' ...

easbar: Yes, and also in JSX: https://github.com/graphhopper/graphhopper-maps/pull/88#discussion_r643045987

karussell: Aha, ok. Probably because of some xml compatibility. Should we use " instead of ' in ts/x?

easbar: I don't know. @Janekdererste ?

Janekdererste: Rather not, but I don't have a strong oppinion on that. 🤷‍♀️

easbar:
I just added the prettier command to `package.json` (to use it do: `npm run format`).

src/index.html
easbar: So let's keep the double quotes for html and tsx and merge?

Janekdererste: I'd vote yes. Since Prettier is supposed to be an oppinionated formatter, we should stick to the defaults as long as nobody has a strong oppinion to do things otherwise.

karussell: :+1:

package.json
```diff
@@ -4,6 +4,7 @@
     "description": "react dummy",
     "main": "index.js",
     "scripts": {
+        "format": "prettier --write \"src/**/*.{css,ts,tsx,html}\"",
```
karussell, line 7: ah, nice

PR #94, opened 2021-06-07
Use readable css module names for dev build

### LINKED ISSUE #73: Do not obfuscate css class names during development
For example in `App.tsx` we use css like this (works because of the `modules: true` setting in `css-loader`?!)

```tsx
<div className={styles.sidebar}>
        <div className={styles.sidebarContent}>
        <Sidebar info={info} query={query} route={route} error={error} />
</div>
```

This way the class names will be obfuscated in dev-tools, which is a bit annoying:

image

Just a minor thing, but maybe it's easy to fix?

easbar:
Fixes #73, the DOM now looks like this:

image

### DIFF
--- a/webpack.dev.js
+++ b/webpack.dev.js
@@ -1,21 +1,48 @@
-const { merge } = require('webpack-merge')
+const { mergeWithRules, CustomizeRule } = require('webpack-merge')
 const path = require('path')
 
 const common = require('./webpack.common.js')
 
-const config = merge(common, {
+const develop = {
     mode: 'development',
     devtool: 'source-map',
     devServer: {
         contentBase: path.resolve(__dirname, 'dist'),
         https: false,
         port: 3000,
-        host: '0.0.0.0',
+        host: '0.0.0.0'
     },
-})
-
-config.module.rules[3].use[1].options.modules = {
-    localIdentName: '[path][name]__[local]'
+    module: {
+        rules: [
+            {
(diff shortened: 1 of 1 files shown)


webpack.dev.js
```diff
@@ -13,3 +13,9 @@ module.exports = merge(common, {
         host: '0.0.0.0',
     },
 })
+
+config.module.rules[3].use[1].options.modules = {
+    localIdentName: '[path][name]__[local]'
```
easbar, line 18: I took this from the documentation here: https://webpack.js.org/loaders/css-loader/#localidentname
Not really sure we absolutely need path and name and we could also use local+hash, but also no real reason not to use the recommended format?

webpack.dev.js
```diff
@@ -13,3 +13,9 @@ module.exports = merge(common, {
         host: '0.0.0.0',
     },
 })
+
+config.module.rules[3].use[1].options.modules = {
```
easbar, line 17: This is certainly a bit brittle and will fail if, e.g. we change the order of the `rules` or `use` arrays in `webpack.common.js`. But the 'merge' command can probably not do this and in case it ever fails we will notice quickly? Another option would be turning the config into a function with arguments? Or maybe even change the `localIdentName` in the common config (also for production)?

easbar: Ah ok I did not see this. I now adjusted the config accordingly.

PR #98, opened 2021-06-08
Do not ignore initial bbox

easbar:
When the map loads and the bounding box is not the entire world (e.g. when using a local GH server with a smaller server) the map should be focused to the bounding box of the available road network. However, without this fix the initial bounding box we get from /info is ignored.

### DIFF
--- a/src/map/Mapbox.ts
+++ b/src/map/Mapbox.ts
@@ -318,14 +318,12 @@ export default class Mapbox {
     }
 
     fitBounds(bbox: Bbox) {
-        if (bbox.every(num => num !== 0)) {
-            this.map.fitBounds(new LngLatBounds(bbox), {
-                padding: Mapbox.getPadding(),
-                duration: 500,
-                animate: !this.isFirstBounds,
-            })
-            if (this.isFirstBounds) this.isFirstBounds = false
-        }
+        this.map.fitBounds(new LngLatBounds(bbox), {
+            padding: Mapbox.getPadding(),
+            duration: 500,
+            animate: !this.isFirstBounds,
+        })
+        if (this.isFirstBounds) this.isFirstBounds = false
     }
 
     getViewPort(): ViewPort {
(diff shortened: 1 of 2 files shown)


src/stores/RouteStore.ts
```diff
@@ -12,7 +12,7 @@ export interface RouteStoreState {
 export default class RouteStore extends Store<RouteStoreState> {
     private static getEmptyPath(): Path {
         return {
-            bbox: [0, 0, 0, 0],
+            bbox: undefined,
```
easbar, line 15: We had this discussion before :) In `BasePath` the bbox is declared optional, but here we set it to `[0, 0, 0, 0]` for the empty path. We then check for `route.selectedPath.bbox` here: https://github.com/graphhopper/graphhopper-maps/blob/3d327b00f95ad4d05c8502777045de1b0038de97/src/App.tsx#L53, but this does not what we want. We would have to change this to `route.selectedPath.bbox.every(num => num !== 0)`, but since bbox is declared optional we even need `route.selectedPath.bbox && route.selectedPath.bbox.every(num => num !== 0)`. So we could do this but then I think bbox should not be optional in `BasePath`. Then again if bbox might be missing I think declaring it optional and handling the cases where it is missing is perfectly fine.

src/map/Mapbox.ts
```diff
@@ -318,14 +318,12 @@ export default class Mapbox {
     }
 
     fitBounds(bbox: Bbox) {
-        if (bbox.every(num => num !== 0)) {
```
easbar, line 321: Here we do not need the `bbox.every` check, because if there was no bounding box it would be undefined and since the type here is `Bbox` (not optional) we do not have to do this check and can rely on the fact it is not missing and we checked for the missing value elsewhere.

src/stores/RouteStore.ts
Janekdererste: Yes, I agree. Since [0,0,0,0] is still a valid bbox having it undefined to indicate we really don't have a bbox is probably better.

src/map/Mapbox.ts
Janekdererste: Yes, as stated in the other comment

Janekdererste (approved):

easbar:
Thanks for taking a look :+1:

Issue #97, opened 2021-06-08
test i18n with various Accept-Language headers


karussell:
- could be minus instead of underscore
 - could be comma separated -> no. This is already handled from the browser
 - fallback to en_US if nothing found
 - could be just `de` and should use `de_DE` (?) or the opposite

easbar:
My browser seems to use `en-US` so with current master the app crashes. We probably should use a fallback translation map in case the locale is unknown. So far we only have a fallback in case the locale was found but some particular string was not found:

https://github.com/graphhopper/graphhopper-maps/blob/3d327b00f95ad4d05c8502777045de1b0038de97/src/translation/Translation.ts#L32-L39

karussell:
Oh. Will fix

PR #99, opened 2021-06-08
avoid crash if navigator.language not found

karussell:
fix for #97


easbar (approved):
Thanks, works for me.

PR #100, opened 2021-06-09
Use pointer events instead of click event to select a geocoding entry

### LINKED ISSUE #91: picking from auto completion is no longer possible
For firefox mobile on Android only. For desktop firefox and chrome on Android it works.

I think this is since #88. 

When picking the location it does not show the marker and also does not know the resolved location.

Janekdererste:
also reduce debounce when typing to 300ms to make the geocoding search feel more responsive

fix #91

@easbar @karussell can you please test this on your devices? I got slightly strange behaviour when changing the order of the Address selected dispatch and the triggered blur on the text input. Just want to make sure this doesn't only accidentially works on my devices.

### DIFF
--- a/src/sidebar/search/GeocodingResult.tsx
+++ b/src/sidebar/search/GeocodingResult.tsx
@@ -41,18 +41,20 @@ const GeocodingEntry = ({
         : styles.selectableGeocodingEntry
     return (
         <li className={styles.geocodingListItem}>
-            {/* This uses pointer down, up and cancel events to make things work on touch devices. Using the click event would close the popup to quickly */}
+            {
             <button
                 className={className}
-                onPointerDown={e => {
-                    setWasCancelled(false)
+                onClick={() => {
+                    console.log('hit selected ' + entry.name)
+                    onSelectHit(entry)
+                }}
+                // Using the click event would close the popup to quickly, see #100
+                onTouchEnd={e => {
+                    onSelectHit(entry)
                     e.preventDefault()
                 }}
-                onPointerUp={e => {
-                    e.preventDefault()
-                    if (!wasCancelled) onSelectHit(entry)
-                }}
-                onPointerCancel={() => setWasCancelled(true)}
+                // prevent blur event for input textbox
+                onPointerDown={e => e.preventDefault()}
             >
                 <div className={styles.geocodingEntry}>
                     <span className={styles.geocodingEntryMain}>{convertToMainText(entry)}</span>
--- a/src/sidebar/search/GeocodingResult.tsx
+++ b/src/sidebar/search/GeocodingResult.tsx
@@ -1,6 +1,6 @@
 import { GeocodingHit } from '@/api/graphhopper'
 import styles from './GeocodingResult.module.css'
-import React from 'react'
+import React, { useState } from 'react'
 
 export default function GeocodingResult({
     hits,
@@ -34,6 +34,8 @@ const GeocodingEntry = ({
     isHighlighted: boolean
     onSelectHit: (hit: GeocodingHit) => void
 }) => {
+    const [wasCancelled, setWasCancelled] = useState(false)
+
     const className = isHighlighted
         ? styles.selectableGeocodingEntry + ' ' + styles.highlightedGeocodingEntry
         : styles.selectableGeocodingEntry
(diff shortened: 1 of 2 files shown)


karussell:
It selects the location, yes :+1: 

There is now the issue that when the location was selected for the start it immediately starts the dialog to select it for the destination and when selecting a suggestion there it immediately opens up the profile selection or the start location again (depending on which item I select). Probably the click event isn't fully consumed or something.

> also reduce debounce when typing to 300ms to make the geocoding search feel more responsive

:+1: 

Could you avoid asking for a suggestion if it is just one character?

src/sidebar/search/GeocodingResult.tsx
```diff
@@ -34,19 +34,24 @@ const GeocodingEntry = ({
...
                 className={className}
-                onClick={() => {
-                    console.log('hit selected ' + entry.name)
-                    onSelectHit(entry)
-                }}
                 // prevent blur event for input textbox
-                onPointerDown={e => e.preventDefault()}
+                onPointerDown={e => {
+                    setWasCancelled(false)
+                    e.preventDefault()
```
karussell, line 49: hm, you already do that ... strange.

Janekdererste: Okay added preventDefault on pointerup as well. Could you please try again? It works for me on Chrome, Firefox (both Desktop and emulated touch device) and iOS 14.6

karussell: Unfortunately not. The same result. I also cleared the cache and restarted the app etc.

Maybe the underlying components get the click before the geocoding selection panel somehow?

easbar:
I tried on two phones using this url: http://gh-maps-react.s3-website.eu-central-1.amazonaws.com/#91_autocomplete_selection_bug and it did not work for firefox and safari on one phone and on the other it worked for chrome but also not for firefox
But maybe it is because of the `#` character in the url? I'll try a different branch name

Janekdererste:
Haha, yes. There are still artifacts on the root folder of the bucket which are from the last executed build from the branches which haven't pulled the deployment script changes from master

Everything afer # is not sent to the server. I guess I have to find a new naming convention for my feature branches.

Janekdererste:
(This branch is not deployed yet)

easbar:
I tried this: http://gh-maps-react.s3-website.eu-central-1.amazonaws.com/issue_100 as well, but same thing :(

karussell:
Having this deployed somewhere to test a branch is really a nice idea :+1: 

btw: I just (re)discovered this easy way to test it locally, so probably you all already know :D ... I wanted to test gh maps with a smartphone and all I had to do was to allow communication of all devices in my wifi and now I am able to live access gh maps at `desktop_ip:3000`. (and if you connect the smartphone with a USB cable you can even remote debug this smartphone - basically close the loop ;) )

src/sidebar/search/GeocodingResult.tsx
karussell: I have no clue what this means and why it works, but instead of your changes I use the original code and only added onTouchEnd, ie. 

```jsx
<button
    className={className}
    onClick={e => {
        onSelectHit(entry)
    }}
    onTouchEnd={e => {
        onSelectHit(entry)
        e.preventDefault()
    }}
    // prevent blur event for input textbox
    onPointerDown={e => e.preventDefault()}
>
```

and this works 🤔 (in firefox desktop, chrome desktop and firefox mobile)

Janekdererste: So, I tested this and also tried mutliple combinations of pointer events and touch and click event. I guess this works then. At least until someone tries to do thing with a stylus :-) 

This will use the click event for mouse interaction and the touch end for touch interaction. I would have found it more elegant to handle everything with pointer events. But I guess we'll leave it like this then.

karussell: > I tested this and also tried mutliple combinations of pointer events and touch and click event. 

Thanks.

I would also keep your changes in AddressInput too as I find it more intuitive to remove focus (blur) as the first step before selecting an element.

> At least until someone tries to do thing with a stylus :-)

You mean these pens?

And yes, it is a bit unsatisfying that touch needs special care, but maybe we understand this better in the future.

I will merge (?)

Janekdererste: > Yes please.

karussell:
Thanks @Janekdererste !

Issue #76, opened 2021-05-19
Api.test.ts does not seem to test anything


easbar:
When I run `npm run test` I get output like this in the console:

```
 PASS  test/routing/Api.test.ts
  ● Console

    console.log
      expect(received).toEqual(expected) // deep equality
      
      - Expected  - 1
      + Received  + 1
      
      @@ -3,10 +3,10 @@
            0,
            0,
            0,
            0,
          ],
      -   "elevation": false,
      +   "elevation": undefined,
          "import_date": "some_date",
          "profiles": Array [],
          "version": "some_version",
        }

      at src/api/Api.ts:58:33
```

However, the tests still pass, so it seems like some expectations are not matched, but the test does not fail.

Also when I add a line like `throw 'something is wrong` in `compareRequestBodyAndResolve` the tests still do not fail

karussell:
To fix the error we just need to add elevation: true to the first test in Api.test.ts:

```
Promise.resolve(
                JSON.stringify({
                    elevation: false,
```

but the underlying problem that the error is not thrown is that the methods infoWithDispatch and routeWithDispatch consume the error via the Promise.catch.

In case of routeWithDispatch we could check for the sent RouteRequestFailed action in the test.
In case of infoWithDispatch we would need to sent a similar action. But probably there is a better solution to test this async stuff.

Janekdererste:
I created a draft PR. The problem was that the api is inherently async. The callback of the api call would trigger the receive method in the test method. 
```
Dispatcher.register({
            receive(action: Action) {
                expect(action instanceof InfoReceived).toBeTruthy()
                expect((action as InfoReceived).result).toEqual(expected)
            },
        })
```
The equality test would fail but the api has a catch clause 
```
infoWithDispatch() {
        this.info()
            .then(result => Dispatcher.dispatch(new InfoReceived(result)))
            .catch(e => console.log(e.message))
    }
```
This catch clause would then handle the error caused by the failed ```expect().toEqual()``` and simply print a message. The test would then finish sucessfully.

Janekdererste:
Now, I've changed things so that

1. The ```Api.infoWithDispatch``` dispatches an ```ErrorAction``` when the catch clause gets called
2. The test now has a more elaborate ```receive``` method i.e.
```
Dispatcher.register({
            receive(action: Action) {
                if (action instanceof InfoReceived) {
                    expect(action.result).toEqual(expected)
                } else if (action instanceof ErrorAction) {
[...]

PR #101, opened 2021-06-09
Fix passing unit tests with errors

Janekdererste:
This PR is related to #76 .

### DIFF
--- a/test/routing/Api.test.ts
+++ b/test/routing/Api.test.ts
@@ -1,17 +1,17 @@
 import fetchMock from 'jest-fetch-mock'
-import Dispatcher, { Action } from '../../src/stores/Dispatcher'
+import Dispatcher from '../../src/stores/Dispatcher'
 import { ErrorAction, InfoReceived, RouteRequestFailed, RouteRequestSuccess } from '../../src/actions/Actions'
 import { ApiImpl, ghKey } from '../../src/api/Api'
-import { ApiInfo, ErrorResponse, RawResult, RoutingArgs, RoutingRequest } from '../../src/api/graphhopper'
+import { ApiInfo, ErrorResponse, RoutingArgs, RoutingRequest } from '../../src/api/graphhopper'
 
 // replace global 'fetch' method by fetchMock
 beforeAll(fetchMock.enableMocks)
 
 // clear everything before each test
-beforeEach(() => fetchMock.mockClear())
-
-// after each test clear the dispatcher in case a dummy store was registered
-afterEach(() => Dispatcher.clear())
+beforeEach(() => {
+    fetchMock.mockClear()
+    jest.clearAllMocks()
+})
 
 // disable fetchMock and restore global 'fetch' method
 afterAll(() => fetchMock.disableMocks())
@@ -28,6 +28,7 @@ describe('info api', () => {
         }
 
         fetchMock.mockResponse(request => {
+            // first assert, that the api is called correctly
             expect(request.method).toEqual('GET')
             expect(request.url.toString()).toEqual(expectedUrl)
             expect(request.headers.get('Accept')).toEqual('application/json')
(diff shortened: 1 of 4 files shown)


test/routing/Api.test.ts
```diff
@@ -39,17 +37,27 @@ describe('info api', () => {
...
                     version: expected.version,
                     profiles: [],
+                    elevation: expected.elevation,
                 })
             )
         })
 
+        // this is a little indirect. we are making assertions in the callback of the async api call
+        // if the assertion is wrong the error is caught by the error handler of the api, dispatching an errorAction
+        // therefore we must make the test fail manually
```
karussell, line 47: Couldn't we instead return the promise to the test and make the assert there?

Janekdererste: There is an async and a synchronous method in the api. We could certainly test the async ones, which would make testing easier. On the other hand our application uses the "withDispatch" methods. We would not test whether the expected actions are dispatched after an api call happens.

easbar: If we want to test the 'dispatch' calls I think the most natural thing would be to spy on `Dispatcher` rather than actually registering listeners, something like this:

```javascript
// our spy will register all calls to Dispatcher.dispatch
const spy = jest.spyOn(Dispatcher, 'dispatch')
// run our code. we cannot await this because infoWithDispatch is not async
new ApiImpl().infoWithDispatch()
// because infoWithDispatch does an async call but does not return a Promise
// we need to make sure we wait until it is finished... 
// I took this from here: https://github.com/facebook/jest/issues/2157#issuecomment-279171856 we could make this a helper/util function
const flushPromises = () => new Promise(setImmediate)
await flushPromises()
// these are all the calls that were made to Dispatcher.dispatch
console.log(spy.mock.calls)
// we can setup some expectations
expect(spy).toHaveBeenCalled()
expect(spy).toHaveBeenCalledTimes(1)
expect((spy.mock.calls[0][0] as any).result).toEqual(expected)
```

Alternatively we could maybe we return a Promise from `infoWithDispatch` (or use an optional callback?), then we would not need `flushPromises` to make sure our `expect` calls run after the `Dispatcher` was called.

Janekdererste: Ah, of course. Spying on the dispatcher is much better! Thanks for the suggestion I'll try to go with this approach. 

> Alternatively we could maybe we return a Promise from infoWithDispatch

The ```infoWithDispatch``` methods are there so that they can be called without ```await``` but with no warnings that promises are not handled. Maybe for the sake of better error handling we could disable this warning and make those methods async as well.

easbar: > The infoWithDispatch methods are there so that they can be called without await but with no warnings that promises are not handled. Maybe for the sake of better error handling we could disable this warning and make those methods async as well.

Or use an optional callback? But unless there are cases where it does not work I think the `flushPromises` thing isn't so bad either.

Janekdererste: Okay, went for the flush promises approach. What do you think now? If you're happy I will merge this.

easbar: Thanks, looks good to me. The only thing I noticed is that still no test fails when I put something like `expect(2).toEqual(3)` in `compareRequestBodyAndResolve`.

Janekdererste: So, when I tested this it worked. The tests fail because the action received by the Dispatcher is not the same as the expected one.

easbar: Ah yes you are right. I did not realize there are multiple `describe` blocks in `Api.test.ts`.

easbar (approved):

test/routing/Api.test.ts
Janekdererste: I will merge this then.

PR #103, opened 2021-06-10
Handle 500 errors from /route

### LINKED ISSUE #70: server side timeout of alternative routes results in JS error

easbar:
Fixes #70 
The issue is not strictly related to alternative routes, but whenever the GH API times out and we receive 500 our code failed because it was assuming there is a `hints` field in the error response that we only get for 400 errors.

### DIFF
--- a/src/api/Api.ts
+++ b/src/api/Api.ts
@@ -14,8 +14,7 @@ import {
     RoutingProfile,
 } from '@/api/graphhopper'
 import { LineString } from 'geojson'
-import { getTranslation } from '@/translation/Translation'
-const tr = getTranslation()
+import { tr } from '@/translation/Translation'
 
 interface ApiProfile {
     name: string
@@ -102,7 +101,7 @@ export class ApiImpl implements Api {
             }
         } else if (response.status === 500) {
             // not always true, but most of the time :)
-            throw new Error(tr.get('route_timed_out'))
+            throw new Error(tr('route_timed_out'))
         } else if (response.status === 400) {
             const errorResult = (await response.json()) as ErrorResponse
             let message = errorResult.message
@@ -112,7 +111,7 @@ export class ApiImpl implements Api {
                     (errorResult.hints as any[]).map(hint => hint.message).join(' and ')
             throw new Error(message)
         } else {
-            throw new Error(tr.get('route_request_failed'))
+            throw new Error(tr('route_request_failed'))
         }
     }
 
--- a/src/api/Api.ts
+++ b/src/api/Api.ts
@@ -14,6 +14,8 @@ import {
     RoutingProfile,
 } from '@/api/graphhopper'
 import { LineString } from 'geojson'
+import { getTranslation } from '@/translation/Translation'
+const tr = getTranslation()
 
 interface ApiProfile {
     name: string
@@ -100,7 +102,7 @@ export class ApiImpl implements Api {
             }
         } else if (response.status === 500) {
             // not always true, but most of the time :)
-            throw new Error('Route calculation timed out')
+            throw new Error(tr.get('route_timed_out'))
         } else if (response.status === 400) {
             const errorResult = (await response.json()) as ErrorResponse
             let message = errorResult.message
@@ -110,7 +112,7 @@ export class ApiImpl implements Api {
                     (errorResult.hints as any[]).map(hint => hint.message).join(' and ')
             throw new Error(message)
         } else {
-            throw new Error('Route request failed')
+            throw new Error(tr.get('route_request_failed'))
         }
     }
 
--- a/src/api/Api.ts
+++ b/src/api/Api.ts
@@ -98,21 +98,19 @@ export class ApiImpl implements Api {
                 ...rawResult,
                 paths: ApiImpl.decodeResult(rawResult, completeRequest.elevation),
             }
-        } else if (response.status !== 400) {
-            console.log(`route request failed, status: '${response.statusText}'`)
-            if (response.status === 500)
-                // not always true, but most of the time :)
-                throw new Error('Route calculation timed out')
-            throw new Error('Route request failed')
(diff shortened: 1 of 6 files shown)


src/api/Api.ts
```diff
@@ -98,6 +98,8 @@ export class ApiImpl implements Api {
                 ...rawResult,
                 paths: ApiImpl.decodeResult(rawResult, completeRequest.elevation),
             }
+        } else if (response.status !== 400) {
+            throw new Error(response.statusText)
```
easbar, line 102: The GraphHopper API returns 500 when a request times out, so when this happens we see `Internal Server Error' for example. Not sure if this is too unfriendly for users and what else we should do. Maybe 'Unexpected Error' or even 'The request timed out'? In most cases 500 should be because the request timed out and even when the reason was something else we could still show this message? And maybe log the actual error to the console?

src/api/Api.ts
```diff
@@ -98,6 +98,12 @@ export class ApiImpl implements Api {
                 ...rawResult,
                 paths: ApiImpl.decodeResult(rawResult, completeRequest.elevation),
             }
+        } else if (response.status !== 400) {
+            console.log(`route request failed, status: '${response.statusText}'`)
+            if (response.status === 500)
+                // not always true, but most of the time :)
+                throw new Error('Route calculation timed out')
+            throw new Error('Route request failed')
```
easbar, line 106: After thinking about this again I decided to use these more user-friendly error messages. WDYT? And should we translate the error messages (already?)

Janekdererste: Since we have translation now, I'd say yes, if we have an appropriate translation.

I find the different paths of the else if block not easy to read. isn't it somethig along the lines like 
```
if (response.ok) {...}
else if (response.status === 500) {
    throw new Error('Route calculation timed out')
} else if (response.status === 400) {
    throw new Error('Insert your meaningful bad request message here')
} else {
    console.log(`route request failed, status: '${response.statusText}'`)
    throw new Error('Route request failed.')
}

**Review by Janekdererste — APPROVED** [review]
Yes! Friendlier user messages.

File: src/api/Api.ts (hunk 2 above)
Comment (karussell) [reply src/api/Api.ts:106]: > if we have an appropriate translation

Everyone can [add translations here](https://docs.google.com/spreadsheets/d/10HKSFmxGVEIO92loVQetVmjXT0qpf3EA2jxuQSSYTdU/edit?pli=1#gid=0) and we can use them now, yes :)

Add text to English at least (and maybe German) and prefix the key with `web.` and then run `python3 update-translations.py`.

Comment (easbar) [reply src/api/Api.ts:106]: I cleaned up the if/else logic and added the translations (this worked right away :+1:)

File: src/api/Api.ts [hunk 3]
```diff
@@ -14,6 +14,8 @@ import {
     RoutingProfile,
 } from '@/api/graphhopper'
 import { LineString } from 'geojson'
+import { getTranslation } from '@/translation/Translation'
+const tr = getTranslation()
```
easbar, line 18: This is not ideal, because just loading the file executes `getTranslation()` and this fails if there was no `setTranslation()` before. Why do we even need `getTranslation()`? Can we not keep the `Translation` object local to `Translation.ts` and export only the `tr` function instead?

karussell: Probably a good idea, yes. I played around and wasn't sure how to integrate this properly to make it convenient but also aligned to other code. My biggest problem with it is that this is a singleton but probably this is fine for a web app :)

easbar: Can you take a look at my last commit to see if you like it?

easbar: And what is your issue with the singleton? I think `tr.json` should only be loaded once (this data is global so to say). Other than that all we need is some handy access function to retrieve the data. We also need to keep some state to keep track which is the current language. This is handled by the `Translation` object currently and it makes sense there is only one such object?

src/translation/Translation.ts
```diff
@@ -12,7 +12,7 @@ export class Translation {
     }
 
     get(key: string, parameters?: string[]): string {
-        var str: string = this.data[key]
+        let str: string = this.data[key]
```
karussell, line 15: ups, I hope I learn this at some point :) (`let` sounds so arbitrary)

src/api/Api.ts
karussell: > And what is your concern about the singleton?

I just dislike singletons in general (probably from my previous life as Android dev where it made everything a mess :D )

> Can you take a look at my last commit to see if you like it?

Looks even more convenient, yes. But the mechanism is not really changed. I.e. if setTranslation is not called before the usage you'll still get an error (?)

src/translation/Translation.ts
easbar: Just use `const` 90% of the time :)

src/api/Api.ts
easbar: > The mechanism is not really changed though. I.e. if setTranslation is not called before the usage you'll still get an error 

Yes. `setTranslation` still needs to be called before using it (makes sense). The problem was that I put `const tr = getTranslation()` at the beginning of the file. I did this because otherwise I would have to do `getTranslation().get(..)` wherever I wanted to translate something. But on the other hand when put `const tr = getTranslation` at the beginning of the file this is executed as soon as the file is loaded and the order of the import statements becomes relevant which I thought was ugly.

karussell: Ah, ok :+1:

karussell:
Nice @easbar !

PR #107, opened 2021-06-16
#91 autocomplete selection bug

Janekdererste:
Also have pointer cancel event to enable scrolling of geocoding result on touch devices


karussell:
Thanks - LGTM!

BTW: Could this be a firefox bug? (as it worked in chromium/chrome I think?)

Janekdererste:
> BTW: Could this original ugliness be a firefox bug? (as it worked in chromium/chrome I think?)

Maybe it is also just different browser behaviour? I din't really find anything when I tried to google for this issue. Also, I found it hard to debug because of which event a textbox receives focus.

PR #108, opened 2021-06-16
Move layers to separate files

easbar:
Based on #87 I tried making our map component independent from the different layers that are added throughout the application.


src/App.tsx
```diff
@@ -54,48 +69,100 @@ export default function App() {
...
+                if (feature.layer.id === pathsLayerKey) {
+                    const index = feature.properties!.index
+                    const path = getCurrentPaths(route.selectedPath, route.routingResult.paths).find(
+                        indexPath => indexPath.index === index
+                    )
+                    Dispatcher.dispatch(new SetSelectedPath(path!.path))
+                }
+            },
+            layer: <PathsLayer selectedPath={route.selectedPath} paths={route.routingResult.paths} />,
+        },
```
easbar, line 129: @Janekdererste maybe you can help me here? Basically I would like to move all this into `PathsLayer.tsx` for example, so here in `App.tsx` it would maybe simply look like this: 

```javascript
const mapLayers = [
  new ContextMenuLayer(popupCoordinate),
  new QueryPointLayer(query.queryPoints),
  new PathsLayer(route.selectedPath, route.routingResult.paths),
  new PathDetailsLayer(pathDetails),
]
mapLayers.forEach(layer => {
  Dispatcher.dispatch(new PutMapLayerAction(layer))
})
```

But I could not figure out how to move the JSX out from here. Note that I already set up the `MapLayerStore` but not using it so far.

src/map/Map.tsx
```diff
@@ -1,51 +1,26 @@
...
 type MapProps = {
     viewport: ViewportStoreState
-    selectedPath: Path
-    paths: Path[]
-    queryPoints: QueryPoint[]
     mapStyle: StyleOption
-    pathDetailPoint: PathDetailsPoint | null
-    highlightedPathDetailSegments: Coordinate[][]
+    mapLayers: { [key: string]: MapLayer }
+    setPopupCoordinate: (c: Coordinate | null) => void
```
easbar, line 15: Passing around the `setPopupCoordinate` function here is against the current architecture and we probably would need another store for this. So far we were able to keep `const [popupCoordinate, setPopupCoordinate] = useState(null)` in `Map.tsx`, but now that `Map` and `ContextMenuLayer` are separate components this is more complicated.

src/App.tsx
easbar: My goal was that adding a layer would be a matter of adding a file that specifies the layer and calling `Dispatch.dispatch(new PutMapLayerAction(layer))` somewhere. The latter is probably done only when the app loads, but in principle we might add/remove layers interactively?! We need the `interactiveLayerIds` and `onClick` propertys for every such layer, because the Mapbox/React wrapper needs to pass the user events to all our layers.

src/map/Map.tsx
easbar: Hm, or we move the `ContextMenuLayer` back into `Map.tsx` because it is kind of more related to the map than the others? :) Btw. the content of the context menu is definitely something that we might need to update based on user-interaction, e.g. when switching between /route and /isochrone mode for example.

src/App.tsx
easbar: Sorry I went back and forth here a bit and now I actually minimized the changes and  think this should be included in #87

src/map/Map.tsx
easbar: ... this is what I did :)

Issue #95, opened 2021-06-08
Send route query again when pressing enter


easbar:
Continuing the discussion from here: https://github.com/graphhopper/graphhopper-maps/pull/88#issuecomment-853754364

> Which behaviour do we want in this case? Simply try to use the text-input's content as coordinate? Only if it has changed since the last time? Or should the Enter -> Search thing be a more global thing? Also, should the search be triggered only in case an input box is highlighted or again at any time?

Yes, I would trigger the routing request if Enter is pressed while the cursor is in one of the input boxes and either do geocoding with nominatim or, if it is a GPS coordinate use it as is (compare this comment: https://github.com/graphhopper/graphhopper-maps/issues/54#issuecomment-856893896). It should not matter whether anything changed since last time, simply send the request or is this somehow complicated? I do not think the Enter key should be global, it should be enough to have it while the focus is in one of the input boxes.

karussell:
Currently this does not work if auto complete box is not yet shown.

PR #112, opened 2021-06-17
Use Tab to select adresses

### LINKED ISSUE #54: leaving input with tab should work (keyboard navigation)

Janekdererste:
fix #54 .

This works slightly different than described in #54 now. I kept the tab order as is, so that the delete buttons next to the input fields are still accessible via tab. Please let me know how this works for you. 

http://gh-maps-react.s3-website.eu-central-1.amazonaws.com/54_tab_search

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -138,6 +138,7 @@ function LargeScreenLayout({ query, route, bbox, error, mapOptions, info, pathDe
                             points={query.queryPoints}
                             routingProfiles={info.profiles}
                             selectedProfile={query.routingProfile}
+                            autofocus={true}
                         />
                     </div>
                     <div>{!error.isDismissed && <ErrorMessage error={error} />}</div>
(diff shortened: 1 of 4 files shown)


karussell:
Nice, this works for me :+1: 

A minor glitch is when you type fast and press "tab" before the suggestions appear - then the location won't resolve (ok if we are able to later press ENTER) and the suggestion list doesn't disappear (this is a bug I think)

karussell:
This should also "somehow" :) work when we enter coordinates: #114

Janekdererste:
> A minor glitch is when you type fast and press "tab" before the suggestions appear

So, I set this to cancel geocoding requests when the input looses focus before the geocoding results are displayed. I think this is what is possible as long as we don't do geocoding requests "on the fly". Let me know if this is ready to merge or whether I missunderstood @karussell

karussell:
Yes, thank you. This is what I meant. (With #114 and #95 this will be the/my expected behaviour.)

PR #120, opened 2021-07-06
move api keys out of the source code

karussell:
To avoid using our API keys in other applications I tried to externalize them using environmental variables.

This works through the EnvironmentPlugin which basically replaces any occurrence of `process.env.XY` with the content of XY plus the string quotation (i.e. JSON.stringify(XY)).

TODO:

 - [x] Is this the correct way to do this? I didn't find a different approach and e.g. deck gl seems to do this similarly. In old gh maps we used browserify-swap which I liked more, but made problems in the last months.
 - [x] hide the API key also in travis
 - [x] use a json file (or something not in git) in the webpack process to avoid setting up the variables everytime we start development (`npm run watch`). This would also allow simpler switching between development and production configs.
 - [x] explain this setup in readme

### DIFF
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -20,7 +20,7 @@ if (!apikeys.graphhopper) {
     console.log('module.exports = ' + JSON.stringify(emptyApiKeys))
     process.exit(-1)
 } else {
-  //  console.log('module.exports = ' + JSON.stringify(apikeys))
+    console.log('module.exports = ' + JSON.stringify(apikeys))
 }
 
 module.exports = {
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -17,10 +17,10 @@ if (!apikeys.graphhopper) {
 
     console.log('Missing apikeys.js file or missing graphhopper key inside this file. Get it from https://graphhopper.com/dashboard')
     console.log('Then create the file with:')
-    console.log('module.exports = ' + JSON.stringify(emptyApiKeys))
+    console.log('module.exports=' + JSON.stringify(emptyApiKeys))
     process.exit(-1)
 } else {
-    console.log('module.exports = ' + JSON.stringify(apikeys))
+    // console.log('module.exports=' + JSON.stringify(apikeys))
 }
 
 module.exports = {
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -1,6 +1,27 @@
 const path = require('path')
+const webpack = require('webpack')
+
 const HTMLWebpackPlugin = require('html-webpack-plugin')
 const FaviconsWebpackPlugin = require('favicons-webpack-plugin')
+
+const apikeys = require('./apikeys.js')
+
+if (!apikeys.graphhopper) {
+    const emptyApiKeys = {
+       "graphhopper": "missing api key",
+       "maptiler": "missing api key",
+       "omniscale": "missing api key",
+       "thunderforest": "missing api key",
+       "kurviger": "missing api key",
+    }
+
+    console.log('Missing apikeys.js file or missing graphhopper key inside this file. Get it from https://graphhopper.com/dashboard')
+    console.log('Then create the file with:')
+    console.log('module.exports = ' + JSON.stringify(emptyApiKeys))
+    process.exit(-1)
+} else {
+  //  console.log('module.exports = ' + JSON.stringify(apikeys))
+}
 
 module.exports = {
     entry: path.resolve(__dirname, 'src', 'index.tsx'),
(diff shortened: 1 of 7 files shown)


Janekdererste:
Putting this into an environment variable is an appropriate approach I think. 

Looking at the files it seems like you have checked in a key. Is this an intermediate step? (Asking because you stated in your initial comment that you don't want the keys in a repo.)

karussell:
Yes, the key is still visible and we could hide it via some travis SECRET. I just wanted to make sure if we should go further in this direction or if there is a better approach in JS-world :)

Will then continue here and let you know once I think this is mergable.

karussell:
Using the same mechanism I have now externalized it into a single JSON file so that no special shell command has to executed before and instead a file `apikeys.js` lives now in the root of the repo but is ignored from git.

For travis this was a bit tricky as I needed to escape the value of the json file like:
```
module.exports=\{\"graphhopper\":\"...
```

webpack.common.js
```diff
@@ -66,6 +87,13 @@ module.exports = {
     plugins: [
         new HTMLWebpackPlugin({ template: path.resolve(__dirname, 'src/index.html') }),
         new FaviconsWebpackPlugin(path.resolve(__dirname, 'src/favicon.png')),
+        new webpack.EnvironmentPlugin({
+            GraphHopperApiKey: apikeys.graphhopper, // use no default to throw exception if missing
```
easbar, line 91: What about using the web app with graphhopper running on localhost? Maybe we should already rename `apikeys.js` to `config.js` and include host and port and rather make the graphhopper key optional?

karussell: Sure we can make these configurable too and rename the file name.

But I'm unsure if we should make the key optional in this UI-only repo as we would then have to explain people on how to set up the graphhopper routing server. Instead I would require the API key and link to graphhopper routing server installation instructions if this is not wanted.

karussell: And I'm not saying this from some suspicious commercial interests ;D ... I mean we should also require MapTiler API key (like deck.gl requires mapbox api keys), because we expect no one to setup OpenMapTiles only to run this UI project.

easbar: Ok and when we want to use it with local GraphHopper we just need to enter a dummy key?

easbar: I wonder who is willing to get an api key just to develop the UI, I'd rather use a local GraphHopper instance and instructions how to set this up can be easily found at the GH repo. And btw don't you want to move the web app into the GH repo anyway? I mean once this replaces the old web app it would be nice to have it out of the box when setting up a GH server?!

karussell: > Ok and when we want to use it with local GraphHopper we just need to enter a dummy key?

Yes, or we detect the host and if it contains localhost we ignore this requirement?

> I'd rather use a local GraphHopper instance and instructions how to set this up can be easily found at the GH repo

IMO if you are a front end dev you don't want to mess with a JVM ;) ... also again, why require a maptiler key and not the GraphHopper key?

> And btw don't you want to move the web app into the GH repo anyway?

Not so sure anymore - what do you think? Having separate repos has the advantage of a different release schedule. And in GH core we could fetch a working release. Of course has also disadvantages ...

> I mean once this replaces the old web app it would be nice to have it out of the box when setting up a GH server?!

Of course, sure.

easbar: As long as it is somehow possible to develop against localhost I'm happy :) Now that we have a config file I'd also wish we can configure the host in this config file, but if not we can still modify `Api.ts`.

> also again, why require a maptiler key and not the GraphHopper key?

I guess this is a frontend to GraphHopper so some interest/knowledge about GraphHopper can be expected I think. Also one can develop using openstreetmap tiles without a key.

> Not so sure anymore - what do you think? Having separate repos has the advantage of a different release schedule. And in GH core we could fetch a working release. Of course has also disadvantages ...

I'm also not sure, but I don't see much reason for change compared to what we are doing now (keep the webapp inside the graphhopper repo). The main requirement is that the webapp is running once we start the GH server, but you are right we can probably fetch a working release for this.

webpack.common.js
```diff
@@ -1,7 +1,28 @@
...
+if (!apikeys.graphhopper) {
+    const emptyApiKeys = {
+       "graphhopper": "missing_api_key",
+       "maptiler": "missing_api_key",
+       "omniscale": "missing_api_key",
+       "thunderforest": "missing_api_key",
+       "kurviger": "missing_api_key",
+    }
+
+    console.log('Missing apikeys.js file or missing graphhopper key inside this file. Get it from https://graphhopper.com/dashboard')
```
easbar, line 18: This does not work, if `apikeys.js` is missing the require call will already fail (and the error message is not as clear), at least this is happened when I tried it.

karussell:
Let's use #123

PR #123, opened 2021-07-13
Move config to external file

easbar:
After I tried #120 I also tried this myself and this is the result...

I think what we want is a config file that has all the configuration we want (api keys, but also host and maybe later a flag for 'enable mvt' etc.). But since we do not want to keep the api keys in VCS it has to be possible to replace the config file. I am now using webpack's resolve/alias feature which allows redirecting import statements to a given path, and this path can be set conditionally in the webpack config. The condition I use is the existence of a `config-local.js` file which is in `.gitignore`. If this file exists it will be included in the bundle by webpack. If it does not exist the default `config.js` file is used, which is included in VCS. Anywhere in our code we can import the config like `import config from 'config'` where `'config'` is an alias for whatever config file we include in the webpack configuration (it might be `config-local.js` or `config.js`).

So what this means for development is that unless we create a `config-local.js` file the default `config.js` file is used and we can setup our own api keys in `config-local.js`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ global:
     - travis encrypt AWS_SECRET_ACCESS_KEY=$S3_KEY --add
 
 script:
-    - export GraphHopperApiKey='fb45b8b2-fdda-4093-ac1a-8b57b4e50add'; export MapTilerApiKey='not required for testing'
+    - echo "$APIKEYSJS_CONTENT" > apikeys.js
     - npm run build-debug && npm run test
 
 # install awscli command line tool to sync files to s3
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ global:
     - travis encrypt AWS_SECRET_ACCESS_KEY=$S3_KEY --add
 
 script:
-    - echo "$APIKEYS_CONTENT" > apikeys.js
+    - echo "$APIKEYS_CONTENT" > config-local.js
     - npm run build-debug && npm run test
 
 # install awscli command line tool to sync files to s3
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ global:
     - travis encrypt AWS_SECRET_ACCESS_KEY=$S3_KEY --add
 
 script:
-    - export GraphHopperApiKey='fb45b8b2-fdda-4093-ac1a-8b57b4e50add'
+    - export GraphHopperApiKey='fb45b8b2-fdda-4093-ac1a-8b57b4e50add'; export MapTilerApiKey='not required for testing'
     - npm run build-debug && npm run test
 
 # install awscli command line tool to sync files to s3
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ global:
     - travis encrypt AWS_SECRET_ACCESS_KEY=$S3_KEY --add
 
 script:
-    - echo "$APIKEYS_CONTENT" > config-local.js
+    - echo "$CONFIG_CONTENT" > config-local.js
     - npm run build-debug && npm run test
 
 # install awscli command line tool to sync files to s3
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ global:
 
 script:
     - echo "$APIKEYS_CONTENT" > apikeys.js
+    - cat apikeys.js
     - npm run build-debug && npm run test
 
 # install awscli command line tool to sync files to s3
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,6 @@ global:
 
 script:
     - echo "$APIKEYS_CONTENT" > apikeys.js
-    - cat apikeys.js
     - npm run build-debug && npm run test
 
 # install awscli command line tool to sync files to s3
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ global:
     - travis encrypt AWS_SECRET_ACCESS_KEY=$S3_KEY --add
 
 script:
-    - echo "$APIKEYSJS_CONTENT" > apikeys.js
+    - echo "$APIKEYS_CONTENT" > apikeys.js
     - npm run build-debug && npm run test
 
 # install awscli command line tool to sync files to s3
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,9 @@ global:
     - travis encrypt AWS_ACCESS_KEY_ID=$S3_KEY_ID --add
     - travis encrypt AWS_SECRET_ACCESS_KEY=$S3_KEY --add
 
-script: npm run build && npm run test
+script:
+    - export GraphHopperApiKey='fb45b8b2-fdda-4093-ac1a-8b57b4e50add'
+    - npm run build && npm run test
 
 # install awscli command line tool to sync files to s3
 before_deploy: pip install --user awscli
--- a/README.md
+++ b/README.md
@@ -1,13 +1,23 @@
 # GraphHopper Maps
 
-This is a responsive UI for the [GraphHopper routing engine](https://github.com/graphhopper/graphhopper) rewritten from scratch. Soon it will replace the [jquery-based maps UI](https://github.com/graphhopper/graphhopper#graphhopper-maps).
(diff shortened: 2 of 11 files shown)


webpack.common.js
```diff
@@ -1,15 +1,31 @@
...
+const defaultConfig = path.resolve(__dirname, 'config.js')
+let config
+if (fs.existsSync(localConfig)) {
+    config = localConfig
+} else if (fs.existsSync(defaultConfig)) {
+    config = defaultConfig
+} else {
+    throw `The config file is missing: ${defaultConfig}`
+}
+
```
easbar, line 17: In #120 there was an additional check that made sure that config includes a gh key, but I'm not sure if this is needed. If there is no api key we currently get a 401 response for graphhopper.com/api/1/info and if we handle this properly we could show a nice 'missing api key' message right in the UI. But maybe we still need to deal with missing tile keys (unless we just use OSM by default, see above). Not checking for the GH key here has the advantage that it is not needed anyway when we set the url to localhost (using a local GH server).

webpack.common.js
```diff
@@ -1,15 +1,31 @@
...
     entry: path.resolve(__dirname, 'src', 'index.tsx'),
     output: {
         path: path.resolve(__dirname, 'dist'),
         filename: 'bundle.js',
     },
     resolve: {
-        alias: { '@': path.resolve(__dirname, 'src') },
+        alias: {
+            '@': path.resolve(__dirname, 'src'),
+            config$: config
```
easbar, line 27: Here we define the webpack alias, see the docs here: https://webpack.js.org/configuration/resolve/#resolvealias

easbar: One advantage of this approach is that the live dev server recognizes changes in the config file, i.e. we can change the api key while the app is running. We have to restart when we want to switch from `config.js` to `config-local.js`, though, because webpack only executes this code at start up.

webpack.common.js
karussell: Why do we need two config files?

karussell (approved):
Ah, cool - I was sure there is a simpler way - thanks!!

webpack.common.js
karussell: Ok, sorry. Just read your main comment... it is clear now.

easbar: Can we make OSM tiles the default (without api key)? Or maybe omniscale like in maps 1.0? It seems kind of ugly to tell everyone to get a maptiler key, but not sure if using OSM without key is ok.

karussell: Currently raster tiles would be ugly due to the flickering issue. In general we can do this, sure. (for production I would still keep using maptiler or mapilion)

karussell: > not sure if using OSM without key is ok.

They do not have any API key protection

src/custom.d.ts
```diff
@@ -3,6 +3,16 @@ declare module '*.svg'
...
+declare module 'config' {
+    const api : string
+    const keys = {
+        graphhopper: string,
+        omniscale: string,
+        maptiler: string,
+        thunderforest: string,
+        kurviger: string
+    }
+}
```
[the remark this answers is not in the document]
easbar: Can we make OSM tiles the default (without api key)? Or maybe omniscale like in maps 1.0? It seems kind of ugly to tell everyone to get a maptiler key, but not sure if using OSM without key is ok.

README.md
```diff
@@ -1,13 +1,15 @@
...
 
 Try it at [graphhopper.com/maps2](https://graphhopper.com/maps2/).
 
 Get started:
 
- * use your GraphHopper API key at src/api/Api.ts
-   or point it to your local GraphHopper server (change apiAddress in this file)
+ * copy config.js to config-local.js and enter your GraphHopper API key for routing (get it at https://www.graphhopper.com)
+   and your MapTiler key for map tiles. You can also change the 'api' field and point it at your local GraphHopper server,
+   e.g. `api: http://localhost:8989/` and similarly you can host your own tiles using e.g. OpenMapTiles
```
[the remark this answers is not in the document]
easbar: Can we make OSM tiles the default (without api key)? Or maybe omniscale like in maps 1.0? It seems kind of ugly to tell everyone to get a maptiler key, but not sure if using OSM without key is ok.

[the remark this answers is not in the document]
karussell: > Having to setup the maptiler key is ugly I think

It is ugly but it is not too bad. E.g. deck.gl does this and maybe we could ask maptiler to get free maptiles from them if we do this :D

webpack.common.js
easbar: The problem seems to be that we get a CORS error when we request /info with a wrong API key :(

image

This makes it hard to show a nice error message.

easbar: I mean sure we could already check if there is an api key in webpack but a) this check will fail unecessarily when we use localhost and b) even when the api key exists it could still be wrong and we'd like to see a useful error message in the url. Nothing major, but would be nice to have.

README.md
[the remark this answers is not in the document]
easbar: Hm but imagine we switch the frontend of the GH server. I think it wouldn't be nice if we had to setup an API key to see the app when we start the GH server and currently we also do not need to do this.

webpack.common.js
karussell: This is an older known issue behind the scenes which we could/should fix

easbar: > Currently raster tiles would be ugly due to the flickering issue. In general we can do this, sure. (for production I would still keep using maptiler or mapilion)

Yes, I mean we could use OSM by default (in `config.js`) and then in the production config (`config-local.js` that we inject via travis) we set it to map tiler.

> They do not have any API key protection

Yes, I was just wondering if it is ok to use, but actually I don't see a reason not to use it as default when we can already use it via the layers menu.

> This is an older known issue behind the scenes which we could/should fix

Ok, I'll open a separate issue for this.

README.md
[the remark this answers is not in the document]
karussell: Yes this wouldn't be nice, so let's provide a config option which map tile service is default and if it is OSM do not require an API key?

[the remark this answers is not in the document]
easbar: Yes, exactly. Please have a look at my last commit.

.travis.yml
```diff
@@ -8,7 +8,9 @@ global:
     - travis encrypt AWS_ACCESS_KEY_ID=$S3_KEY_ID --add
     - travis encrypt AWS_SECRET_ACCESS_KEY=$S3_KEY --add
 
-script: npm run build-debug && npm run test
+script:
+    - echo "$APIKEYS_CONTENT" > config-local.js
```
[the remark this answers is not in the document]
easbar: @karussell  can you adjust this? Currently the build fails, probably because we create an empty file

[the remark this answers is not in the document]
easbar: `$APIKEYS_CONTENT` should be the same as `config.js`, except that it contains the correct api keys

karussell:
I fixed the config.js for travis but it seems the "config" import in Api.ts cannot be found when running the tests.

easbar:
> I fixed the config.js for travis but it seems the "config" import in Api.ts cannot be found when running the tests.

Oh yes, we have to define such an alias for 'config' also for the test (jest) config. I did this in the last commit. I also had to remove the ES6 export in config.js, so you need to adjust this for the travis variable also (use `module.exports = ` instead of `export default`). jest cannot handle ES6 modules apparently. Another option would probably be to include config.js in the typescript compilation but I did not know how to do this and since it is working now I think it's fine.

jest.config.js
```diff
@@ -1,5 +1,8 @@
 module.exports = {
     preset: 'ts-jest',
     testEnvironment: 'node',
-    moduleNameMapper: { '@/(.*)$': '<rootDir>/src/$1' },
+    moduleNameMapper: {
+        '@/(.*)$': '<rootDir>/src/$1',
+        'config': '<rootDir>/config.js'
```
karussell, line 6: Oh, I was soo close :D ... I just didn't understand that I had to do the change in config.js too.

easbar: Ah :D but thanks for trying

karussell:
Thanks, have adapted the CONFIG_CONTENT now in travis accordingly and triggered the last build again.

PR #125, opened 2021-07-19
Fix infinite loop when back button is pressed

### LINKED ISSUE #116: back button infinite loop
If you go to e.g. [gh maps2](https://graphhopper.com/maps2/?point=49.05176%2C15.782237) in the same browser page and enter a destination you won't be able to return to github using the back button as some automatic redirect is happening. (tested in firefox mobile and desktop)

Janekdererste:
fixes #116 

This needs fixing of unit tests before it can be merged.


easbar:
There are lots of whitespace changes in package-lock.json. When I auto-format the file in IntelliJ most changes go away. Should we apply the formatting? And maybe we should update package-lock.json in master first (unrelated to this issue).

Janekdererste:
Hm, I added .json files to my prettier config in intellij. I didn't think about the package.lock file. Do you think I shoulld revert this?

easbar:
> Hm, I added .json files to my prettier config in intellij. I didn't think about the package.lock file. Do you think I shoulld revert this?

Ah ok so we have to prettier package.lock after creating it. This sounds good to me.

easbar:
For some reason `prettier --write "package-lock.json"` does not change the file? Doing the same for `package.json` works!? In any case the file we commit should be consistent with the `npm run format` script (currently not including `package-lock.json`).

easbar:
Ah `package-lock.json` is in `.prettierignore`...

Janekdererste:
Alright, I took the package-lock.json from master to avoid all the unecessary formatting diff. I think we shold let the lock file be ignored by prettier. This way we can stick with the formatting of the genrated file.

easbar:
Ok thanks. But don't we have to update the package-lock file in master? I thought the build did not pass with the current one?!

Janekdererste:
The build on master was passing. My freshly installed computer generated a new lock file when installing dependencies. I hope that this works like this now.

karussell:
Thanks a lot @Janekdererste !

Issue #122, opened 2021-07-08
dragging or zooming a raster map like OpenStreetMap hides the routes


karussell:
This is a strange bug as it works for all vector tiles.

easbar:
Apparently this was introduced by #87

karussell:
Strange, for "Maptiler Satellite" this works fine although I think that these are raster tiles too (?)

easbar:
Ok, I'll try to create a minimal example for this.

easbar:
https://github.com/visgl/react-map-gl/issues/1530

easbar:
The problem is here: https://github.com/graphhopper/graphhopper-maps/blob/3f7e9830a540875bd098d3d6efe5631121298044/src/map/Map.tsx#L29

With each call of `Map#render` we create a new style object and the map is draw again which leads to the flickering. We get the `mapStyle` option from the `MapOptionsStore` and this is just a constant so it will always be the same object for a given style. However, in `mapStyle` we wrap these constant objects into new ones that are created over and over again.

@Janekdererste any idea how we should fix this? I guess we could 'cache' the objects we produce in `getStyle`, but that does not seem like such a great idea. Or we create the ready to be used (MapBox) mapStyle objects right in the `MapOptionsStore` once and for all. The code kind of looks like you were trying to make the data in `MapOptionsStore` not related to Mapbox or why do we even have this `StyleOption` interface? Otherwise I think I would rather create an interface for the Mapbox style (one for vector and one for raster) and there would be nothing we need to wrap in the first place?
Or maybe react has a some built-in feature that allows 'transforming' a state object (like `mapStyle`) without re-creating new objects all the time? Maybe `React.memo()`?

Janekdererste:
Yes, the code in ```getStyle()``` was copied from the former Mapbox class where it was only run once when the map was created. The code in ```MapOptionsStore``` is supposed to be independent of Mapbox.  I felt like the stores should be independent of the visualization details and the view components should transform our custom data model into whatever they needed to cater their APIs. 

I don't insist on keeping it this way. 

1. Setting Mapbox Layers in the ```MapOptionsStore``` sounds pragmatic and would solve the re-rendering issue. 
2. Another way would be to have the Mapbox style in a state, which is only updated when the Style store changes. I think that is your suggestion from the last paragraph
```
const [style, setStyle = useState(getStyle(mapStyle))
useState(() => setStyle(getStyle(mapStyle)), [mapStyle])
```

[...]

Issue #109, opened 2021-06-17
improve colors


karussell:
@Janekdererste mentioned this: the green in the elevation widget does not fit with the "less aggressive" colors elsewhere and we should use 
![image](https://user-images.githubusercontent.com/129644/122361762-d7873980-cf57-11eb-83fc-0a1be0f602bf.png)

Also the color for the routes should be a bit transparent to 
1. allow reading the streetname and
2. knowing the road class. 

btw: Qwant + Google Maps draw all labels above the route (because they can assume to always have vector tiles)

easbar:
Putting the labels on top of the route layer works like this:

https://github.com/easbar/gh_maps_experiments/commit/ba7a0bc3f980d0d168714ffaa7601a5b3d1745cf

easbar:
> the green in the elevation widget does not fit with the "less aggressive" colors elsewhere and we should use

The green we use currently is the one taken from graphhopper.com/maps, but I changed it to the one from the 'from' marker now. IMO it now looks a bit dark/brownish though :) c7dca1988a10f7e08b2db9541d4cc326c419ca8f

karussell:
Ah, turns out this is already fixed and was a caching issue - nice :)

easbar:
I tried moving the map labels to the top here: 5e2b89aaa9610223fe398b9ce196185f67e2476b
But not really sure yet what to do when the background layer is changed

easbar:
Also I think our current choice of color for unselected routes is not ideal because it is too similar to the label colors and makes them hard to read

![image](https://user-images.githubusercontent.com/17603532/122942419-7d7edd80-d376-11eb-8522-ff6f6c72ae1c.png)

karussell:
It is not too bad IMO. Sure, the readability of the "L1074" label could probably a bit increased when surrounded with white somehow like it is done for the "Uder" label but you can also just switch the selected route.

PR #126, opened 2021-07-20
Show vector tile labels on top of the path layer

easbar:
See #109

The Mapbox style consists of many layers with different types. Layers with type 'symbol' should be shown on top of all other layers including our route path layer. Otherwise street labels etc. will be hidden or hard to read. Mapbox offers a 'beforeId' option that allows specifying the position of custom map layers. So basically all we need to do is setting `beforeId` to the first symbol layer of the currently used map style. The main issue with this is that so far we only pass the url of the map style file to Mapbox which is then loaded by Mapbox internally. Therefore we cannot easily figure out which one is the first symbol layer.

In this PR I therefore:

* load all (three to be exact) Mapbox styles when the app starts and keep them in `MapOptionsStore`
* extract the first symbol layer from the loaded style
* use the id of this layer as `beforeId` when creating the paths layer

I think keeping the Mapbox style in `MapOptionsStore` is fine (I actually think it is better) rather than trying to 'hide' Mapbox related stuff from our state store. Just the fact alone that we are dealing with urls pointing to *Mapbox* styles is Mapbox specific anyway. This way we also get rid of the workaround introduced for #122 and there is no need to deal with things like `url: string | string[]` where `url` could be pointing to a mapbox style or might be a list of tile urls.

Loading the style files eagerly when the app starts is just one way this can be done. We could also do this lazily or every time the map style changes (like we currently do), but it is quite fast anyway, and pretty much negligible compared to the loading times for the different tiles:

image

todo: 

* [ ] currently there are some ugly errors in the console like this: (I don't really understand why)
```
// `water_name_line` is the first symbol layer
mapbox.js:30 Error: Layer with id "water_name_line" does not exist on this map.
```
   ~~Probably related to this: The labels are hidden until we change the map layer for the first time~~ (fixed)
* [ ] Are there any downsides? Does it make sense to fetch the mapbox style manually? Or should we even include it in the bundle?
* [ ] Reducing the opacity of the path layer might still be necessary:
image

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -70,7 +70,7 @@ export default function App() {
 
     const mapLayers: MapLayer[] = [
         createQueryPointsLayer(query.queryPoints),
-        createPathsLayer(route.selectedPath, route.routingResult.paths),
+        createPathsLayer(route.selectedPath, route.routingResult.paths, mapOptions.firstSymbolLayerId),
     ]
     if (!isSmallScreen) mapLayers.push(createPathDetailsLayer(pathDetails))
 
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -118,7 +118,7 @@ function LargeScreenLayout({ query, route, viewport, mapLayers, error, mapOption
                 {
                     <MapComponent
                         viewport={viewport}
-                        styleOption={mapOptions.selectedStyle}
+                        mapStyle={mapOptions.selectedMapStyle}
                         queryPoints={query.queryPoints}
                         mapLayers={mapLayers}
                     />
@@ -164,7 +164,7 @@ function SmallScreenLayout({ query, route, viewport, mapLayers, error, mapOption
                 <MapComponent
                     viewport={viewport}
                     queryPoints={query.queryPoints}
-                    styleOption={mapOptions.selectedStyle}
+                    mapStyle={mapOptions.selectedMapStyle}
                     mapLayers={mapLayers}
                 />
             </div>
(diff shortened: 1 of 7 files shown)


easbar:
> Layer with id "water_name_line" does not exist on this map.

I think this happens because we set the `beforeId` of e.g. the paths layer to `water_name_line` before the background layer has finished loading. So maybe we need to wait for something like Mapbox's `styledata` event and set the `beforeId` only afterwards.

Janekdererste:
> So maybe we need to wait for something like Mapbox's styledata event and set the beforeId only afterwards

At least when the style is changed with a url one has to wait for the styledata callback anyway. Does it cause problems to set the bofereId at this point? When prefetching the styles. This callback should be called pretty much emediately?

Janekdererste:
> The labels are hidden until we change the map layer for the first time

Do you know why?

easbar:
> Do you know why?

This was just an error on my side and I already fixed it here: https://github.com/graphhopper/graphhopper-maps/pull/126/commits/e92640892162aa11c824f93f781cf133accab9ac

> Does it cause problems to set the bofereId at this point? When prefetching the styles. This callback should be called pretty much emediately?

There will be a console error, but otherwise it works, because on the next re-render the `beforeId` will be taken into account.

Janekdererste:
> next re-render the beforeId will be taken into account.
If I understand this correctly, our layers become part of the style object once they have been added to mapbox. Would it make sense to make the path layer a part of the style which is stored in the store? Then, the layer would be 'already there' when the style is loaded and we would only have to modify the data with which the layer is rendered. 

I certainly have no idea how one would do that, I just wanted to propose in case it could be done.

easbar:
I created a minimal example and asked for support upstream: https://github.com/visgl/react-map-gl/discussions/1541

easbar:
Since we focus on raster tiles and will switch to OpenLayers I'll close here.

Issue #118, opened 2021-06-23
Store some state in the URL


karussell:
E.g. for a cycle route it would be nice to select the map layer like it is possible for the old maps with the `layer` parameter:

https://graphhopper.com/maps/?point=51.47212%2C5.970732&point=51.536086%2C5.987549&layer=OpenStreetMap

Then the `key` and `locale` should be picked from the URL but then removed (i.e. not synchronized).

PR #130, opened 2021-09-01
allow using API key from URL

karussell:
Heavy users are blocked. If they have an own API key we should make it possible to let them use it via the URL parameter.
E.g. https://graphhopper.com/maps2/?key=YOUR_KEY

The changes are a bit involved as we currently create two APIImpl instances due to convenience (I think) which is now no longer possible.

Related to #118


karussell:
> The changes are a bit involved

Found a simpler way and used the singleton directly where the apiImpl is used.

Janekdererste:
If you intend to use the api as a singleton like object. We could supply the api key as constructor parameter since it will never change anyway.

Janekdererste:
Changed api to be a real singleton. If you like it as well, you may merge the pr.

karussell:
Thanks a lot!

PR #131, opened 2021-09-01
store layer in URL and pick locale for initial state

### LINKED ISSUE #118: Store some state in the URL

karussell:
This should fix #118


src/NavBar.ts
```diff
@@ -1,35 +1,40 @@
...
 
-    private onQueryStoreChanged() {
+    private onQueryStateChanged() {
         if (this.isIgnoreQueryStoreUpdates) return
 
         const newHref = QueryUrl.createUrl(
             this.appContext.location.origin + this.appContext.location.pathname,
-            this.queryStore.state
+            this.queryStore.state,
+            this.mapStore.state
```
karussell, line 37: Not sure where the "root" information of the map layer name should be. I kept it in the MapOptionsStore.

Janekdererste:
This looks good. Although it would result in a conflict with https://github.com/graphhopper/graphhopper-maps/pull/125. I'd rather merge 125 before this one and add the layer information to the new NavBar class

karussell:
Sure :+1:

src/NavBar.ts
Janekdererste: I think I now know what you mean and I think it belongs into the ```MapOptionsStore```

Janekdererste:
Merging it, since the #125 is merged and I resolved the conflicts

PR #137, opened 2021-10-05
106 numbered markers

### LINKED ISSUE #106: Numbered markers
When there are multiple via-points it would be useful to display numbered markers like on graphhopper.com/maps:

image

Another problem is that react-map-gl does not include the default marker we are using currently, so we need to add some kind of markers for #87 anyway. Currently we use a solution based on *.png files, but is this the right way to go or should we use svgs? https://github.com/graphhopper/graphhopper/blob/master/web-bundle/src/main/resources/com/graphhopper/maps/js/lib/leaflet_numbered_markers.js

For our route editor we use different markers as well: 

image

But I think the requirements are kind of the same: We need different colors and it should be possible to display a number. Right, @karussell?

@Janekdererste any thoughts about this?

Janekdererste:
fixes #106


karussell:
This is really great, thanks a lot!

I'm currently testing it locally:

 * when I load [this route](http://localhost:3000/?point=51.334%2C12.411804&point=51.493651%2C12.456395&point=51.522567%2C13.417151&point=51.569757%2C12.744793&point=51.43331%2C14.250083&profile=car&layer=MapTiler) I get a short error message "errorResults.hints undefined" in the panel (and multiple times in the logs: `TypeError: errorResult.hints is undefined`). And sometimes the error message even stays.
 * it seems I cannot add more than 3 via points via right click: it is grayed out
 * when I add too many via points I can no longer access the remaining buttons of the sidepanel - the locations should get a scrollbar when screen is too small or something

Janekdererste:
> when I add too many via points I can no longer access the remaining buttons of the sidepanel - the locations should get a scrollbar when screen is too small (?)

Tried to fix the other two points. This one is not so trivial, since the search input and the route results live in the same panel. At the moment the search input has a fixed (intrinsic) size and pushes the route results to the bottom. The route results are scrollable then. Have to think about a good solution for this.

Janekdererste:
And I have to fix some unit tests of course 🙄

karussell:
> At the moment the search input has a fixed (intrinsic) size and pushes the route results to the bottom. The route results are scrollable then.

Ah, I see. And could we give the points a maximum height of 50% or e.g. 5 via points and then after this it should start scrolling and stop pushing down the route result?

Janekdererste:
>  And could we give the points a maximum height of 50% or e.g. 5 via points

Sounds good.

easbar:
Very nice :+1:

PR #144, opened 2021-10-14
Added fake query param to navi

UnrealValentin:
Added a Query parameter called "fake" to the navi branch to let the fake GPS be toggleable by a query parameter instead of editing the sourcecde. It can be used like this: 
http://127.0.0.1:3000/?point=51.385074%2C14.243598&point=51.43331%2C14.250083&profile=car&layer=OpenStreetMap&fake=69
but also like this:
http://127.0.0.1:3000/?point=51.385074%2C14.243598&point=51.43331%2C14.250083&profile=car&layer=OpenStreetMap&fake
### It does not matter if the query parameter has a value or not.


UnrealValentin:
PS: Above travis error has nothing to do with my pull request it seems and "it works on my computer"™.

src/sidebar/RoutingResults.tsx
```diff
@@ -59,6 +62,22 @@ function RoutingResult({ path, isSelected }: { path: Path; isSelected: boolean }
...
+    var vars = query.split("&");
+    console.log(vars)
+    for (var i=0;i<vars.length;i++) {
+        console.log(vars[i]) 
+        if(vars[i].includes("fake")) {
+            console.log("found fake")
+            return true;
+        }
+    }
+    return(false);
```
karussell, line 78: We do not need to manually parse the url. We can use a URL object instead.

Furthermore it is important to remove all unnecessary logging from production code.

karussell:
> PS: Above travis error has nothing to do with my pull request it seems and "it works on my computer"™.

Did you run the test suite? It seems to be related to the new import of RoutingResults.

src/NavBar.ts
```diff
@@ -7,6 +7,8 @@ import { window } from '@/Window'
 import QueryStore, { QueryPoint, QueryPointType, QueryStoreState } from '@/stores/QueryStore'
 import MapOptionsStore, { MapOptionsStoreState, StyleOption } from './stores/MapOptionsStore'
 
+let fakeNavi: boolean = false
+
```
UnrealValentin, line 11: Export was a useless leftover mistake

karussell: Sorry, I didn't mean the visibility. I mean the nature that an unrelated call of parseURL could change fakeNavi. I know it is rather nit-picky but this will make it easier to get the experimental navi branch or parts of it merged into master.

UnrealValentin:
Changed the check Tod directly tappen in the if statement

karussell:
Can you fix the failing test and replace location.toString with `new URL(window.location.href)`?

karussell:
Thanks!

Issue #147, opened 2021-10-29
test deployment seems to be broken


karussell:
"URL constructor: undefinedinfo is not a valid URL."

http://gh-maps-react.s3-website.eu-central-1.amazonaws.com/master/?point=50.815451%2C9.659668&point=51.271338%2C11.131836

Maybe I messed this up when changing the credentials or the config or something?

Issue #111, opened 2021-06-17
Long press handler does not work as expected


easbar:
On ios12 long-pressing the map shows me the context menu, but as soon as I let go it disappears again, no matter how long I press. I tested this on Safari and Firefox.

On Android it worked in Chrome but on Firefox the behavior was not ideal either: Long pressing showed me the context menu but it would only stay if I kept pressing for a few seconds after the context menu appeared for the first time.

karussell:
Can you include the version for each browser? E.g. my mobile firefox 88.1.1 (build 2015 812 947) works as expected.

easbar:
Oh sure, I added the version information.

easbar:
On the same Android phone the long press worked after upgrading Firefox to 89.1.1

easbar:
On iOS 14.8.1 I cannot open the context menu at all (tried FF, Chrome & Safari). There is a similar issue for Leaflet, but this is about iOS 13, so not sure if it is related: https://github.com/Leaflet/Leaflet/issues/6817

easbar:
I was able to fix this on the OpenLayers branch (ol_exp). This version works on my phone: cb3e74ec91cf0360417a0806919f1204ab315f2a
(it also includes #149)

easbar:
> On iOS 14.8.1 I cannot open the context menu at all (tried FF, Chrome & Safari)

I just figured out that this must be related to iPhone 3D Touch. I can open the context menu with a (very) soft touch, but a harder/normal long touch does not open it.

PR #149, opened 2021-10-29
disable user selection on ios

### LINKED ISSUE #93: Long press in iOS Marks context Menu entries

Janekdererste:
fixes #93 

The only way I was able to disable the user selection on iOS when the map has a long press to bring up the context menu was to set ```-webkit-user-select: none``` on the app-wrapper level. I tried to enable selection for the routing results and the sidebar, but somehow performing a long press on the map always selects things in the search box. 

I would assume that copying text from this application on a phone is less important than random selects of elements in the UI.

### DIFF
--- a/src/App.module.css
+++ b/src/App.module.css
@@ -8,6 +8,7 @@
     grid-template-rows: 100%;
     justify-content: space-between;
     overflow: auto;
+    -webkit-user-select: none;
 }
 
 .map {
@@ -39,7 +40,6 @@
 
 .routingResult {
     overflow: auto;
-    min-height: 5rem;
 }
 
 .mapOptions {


easbar:
Unfortunately this is still not working for me. I am using iOS 14.8.1 and tried both Firefox and Safari. A long-press on the map no longer selects the marker symbol next to the first search box (as in master), but now it does not do anything at all.

Janekdererste:
> but now it does not do anything at all.

This is of course not intended as well ☹️

easbar:
Or is this related to #152?

karussell:
Will re-trigger the build. See #147

easbar:
> Or is this related to #152?

Apparently not

easbar:
As I mentioned in #111 this version (that includes this PR): https://github.com/graphhopper/graphhopper-maps/commit/cb3e74ec91cf0360417a0806919f1204ab315f2a on the ol_exp branch works fine under iOS for me. I just also verified that **without** the present PR I can reproduce #93 on that branch (context menu opens, but some text is selected). So while there still seems to be another issue (that is only fixed on the ol_exp branch currently) this PR is required to fix #93. So we should merge this.

Issue #37, opened 2021-03-23
improvements towards via points


karussell:
* numbers in markers
 * drag and drop of the points in the left panel
 * create via points via dragging from route

PR #153, opened 2021-11-10
Insert via points at closest road segment

easbar:
Currently new via points that are added using the context menu are added as last (just before the destination), no matter where we click. This PR inserts new via points based on the click location. More precisely, we insert via points between the two way points that are connected by the road segment that comes closest to the click location.

I ported this functionality form GH maps 1 (see https://github.com/graphhopper/graphhopper/pull/791, https://github.com/graphhopper/graphhopper/pull/795, [code](https://github.com/graphhopper/graphhopper/blob/master/web-bundle/src/main/resources/com/graphhopper/maps/js/routeManipulation.js)).

This is part of #37

### DIFF
--- a/src/map/Popup.tsx
+++ b/src/map/Popup.tsx
@@ -48,7 +48,9 @@ export function PopupComponent({
                     }),
                 }
             })
-            // we silently assume the way points are the same for all routes (and they are)
+            // note that we can use the index returned by findNextWayPoint no matter which route alternative was found
+            // to be closest to the clicked location, because for every route the n-th snapped_waypoint corresponds to
+            // the n-th query point
             const index = findNextWayPoint(routes, coordinate).nextWayPoint
             Dispatcher.dispatch(new AddPoint(index, coordinate, true))
         }
--- a/src/map/Popup.tsx
+++ b/src/map/Popup.tsx
@@ -4,14 +4,18 @@ import styles from './Popup.module.css'
 import { Coordinate, QueryPoint, QueryPointType } from '@/stores/QueryStore'
 import Dispatcher from '@/stores/Dispatcher'
 import { AddPoint, SetPoint, SetViewportToPoint } from '@/actions/Actions'
+import { RouteStoreState } from '@/stores/RouteStore'
+import { findNextWayPoint } from '@/map/findNextWayPoint'
 
 export function PopupComponent({
     coordinate,
     queryPoints,
+    route,
     onSelect,
 }: {
     coordinate: Coordinate
     queryPoints: QueryPoint[]
+    route: RouteStoreState
     onSelect: () => void
 }) {
     const dispatchSetPoint = function (point: QueryPoint, coordinate: Coordinate) {
@@ -26,7 +30,7 @@ export function PopupComponent({
         )
     }
 
-    const setViaPoint = function (points: QueryPoint[]) {
+    const setViaPoint = function (points: QueryPoint[], route: RouteStoreState) {
         const viaPoints = points.filter(point => point.type === QueryPointType.Via)
         const point = viaPoints.find(point => !point.isInitialized)
         onSelect()
@@ -34,7 +38,19 @@ export function PopupComponent({
         if (point) {
             dispatchSetPoint(point, coordinate)
         } else {
-            Dispatcher.dispatch(new AddPoint(viaPoints.length + 1, coordinate, true))
+            const routes = route.routingResult.paths.map(p => {
+                return {
+                    coordinates: p.points.coordinates.map(pos => {
+                        return { lat: pos[1], lng: pos[0] }
+                    }),
+                    wayPoints: p.snapped_waypoints.coordinates.map(pos => {
+                        return { lat: pos[1], lng: pos[0] }
+                    }),
+                }
+            })
+            // we silently assume the way points are the same for all routes (and they are)
+            const index = findNextWayPoint(routes, coordinate).nextWayPoint
+            Dispatcher.dispatch(new AddPoint(index, coordinate, true))
         }
     }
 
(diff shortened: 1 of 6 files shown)


karussell:
Nice, thank you also for this fix!

src/map/findNextWayPoint.ts
```diff
@@ -0,0 +1,67 @@
+import { Coordinate } from '@/stores/QueryStore'
+
+/**
+ * Finds the way-point that follows the part of a route that is closest to a given location
+ *
+ * @param routes one or more routes to be considered
+ * @param location the location for which the closest route should be found
+ */
+export function findNextWayPoint(
```
karussell, line 9: Can we have a utility class and also export calcDist? (I need/have this for the navi branch already)

src/map/Popup.tsx
```diff
@@ -26,15 +30,27 @@ export function PopupComponent({
...
+                return {
+                    coordinates: p.points.coordinates.map(pos => {
+                        return { lat: pos[1], lng: pos[0] }
+                    }),
+                    wayPoints: p.snapped_waypoints.coordinates.map(pos => {
+                        return { lat: pos[1], lng: pos[0] }
+                    }),
+                }
+            })
+            // we silently assume the way points are the same for all routes (and they are)
```
karussell, line 51: Aren't we always using the way points for the correct route? (Also there should be only one route and no alternative when using one or more via-points. Or maybe I do not understand what you mean here :) )

karussell (changes requested):

src/map/findNextWayPoint.ts
easbar: Sure, will do.

src/map/Popup.tsx
easbar: I was thinking along this line:

* the GH API returns one or multiple paths (if there are alternative routes)
* each of the paths as the `snapped_waypoints` property, even though there is only one set of waypoints (and it is the same for all alternative routes)
* findNextWayPoint finds the index of the next way point for the route that is closest to the clicked location and it returns the index of the route alternative as well
* (only) because the way points are the same for all route alternatives can we use this way point index without even checking which route alternative is the closest to the clicked location

So basically there is a mismatch between the UI (one list of way points) and the API (a list of way points for each route alternative). Since we know that the way points are the same for all alternatives we can just ignore this, that is all I wanted to say.

> Aren't we always using the way points for the correct route? 

Yes, they are automatically correct, because they are the same for each route alternative.

> Also there should be only one route and no alternative when using one or more via-points

Also yes, when there are alternatives there can only ever be two way points, because as soon as there are via points we do not calculate alternatives at all (at least for now).

karussell: Thanks! I think, I got it. More complicated than I thought :)

So this feature also introduces a certain dependency between query points and the route(s).

> So basically there is a mismatch between the UI (one list of way points) and the API (a list of way points for each route alternative)

Just for my understanding: will the assumption get wrong if e.g. the same query points snap to different waypoints? (e.g. when we implement the via point routing via map matching this could indeed happen)

karussell: I think I now better understand what findNextWayPoint does :). But I still do not understand this code comment:

 > we silently assume the way points are the same for all routes (and they are)
 
Do you mean the count of the snapped_waypoints must be the same? Because the algorithm should even work for when B snaps to different snapped_waypoints b and b' in the alternative route (?) and so we don't need to assume that "the [snapped] way points are the same for all routes".

easbar: > Because the algorithm should even work when B snaps to a different snapped_waypoints b' in the alternative route (?) and so we don't need to assume that "the [snapped] waypoints are the same for all routes".

Yes, this is true. I probably rather meant "we silently assume all routes were calculated for the same query points". And we assume this even without this feature, because there is only one list of query points in the first place. I was just trying to say that we get a list of alternative routes from the server, and each one is carrying a list of 'snapped_waypoints', but these way points are not independent from each other. They are based on the same query points. And the count is also the same (exactly one snapped point for each query point). I did not intend to emphasize that the snapped points are the same for each route, but rather that there is a 1:1 correspondence between each list of snapped waypoints and the query points, and that this relation is the same for all lists of snapped waypoints. 

Or to put it another way. Imagine I gave you three lists of coordinates (just routes) and a list of `snapped_waypoints` for each of them. And then I would tell you that the clicked location was between index 3 and 4 of the second of the routes. If you wanted to associate these indices to the query points without taking into account that the second route was the closest you would have to know that index 3 and 4 of the snapped_waypoints list correspond to query point 3 and 4 for all routes. 
 
 But you are right there could be different snapped points for each route and we could still simply take the index from `findNextWayPoint` to determine the **query** points between which we introduce the new point. I'll change the comment to "note that we can use the index returned by `findNextWayPoint` no matter which route alternative was found to be closest to the clicked location, because for every route the n-th snapped_waypoint corresponds to the n-th query point". Does this make it more clear?
 
 This discussion makes this matter sound a lot more complicated than it actually ist :D

karussell: > I'll change the comment to "note that we can use the index returned by findNextWayPoint no matter which route alternative was found to be closest to the clicked location, because for every route the n-th snapped_waypoint corresponds to the n-th query point". Does this make it more clear?

Yes, thank you :)

src/map/Popup.tsx
```diff
@@ -4,14 +4,18 @@ import styles from './Popup.module.css'
...
 
 export function PopupComponent({
     coordinate,
     queryPoints,
+    route,
     onSelect,
 }: {
     coordinate: Coordinate
     queryPoints: QueryPoint[]
+    route: RouteStoreState
```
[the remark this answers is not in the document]
Janekdererste: You could move that locig into the store, but I don't know whether this would improve things. I guess you could also pass a subset of properties instead of the whole state, but this would not be such a big difference I guess.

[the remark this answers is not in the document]
easbar: Ok, let's merge this then so we can move on :)

Issue #134, opened 2021-10-01
Flickering/vanishing elevation detail overlay


otbutz:
![grafik](https://user-images.githubusercontent.com/22315436/135608293-dab8f60d-235a-4729-b7ad-b80b6806c4a8.png)

The overlay often flickers/vanishes if i move the cursors along the graph. Leaving the colored area or if the overlay is flipped(black line moves to the right side of the overlay) seems like the main culprit.

Tested on Firefox 92.

easbar:
> The overlay often flickers/vanishes if i move the cursors along the graph

Yes, this seems to happen when the cursor leaves the background (white) area and enters the colored/graph area (or the other way around)

> or if the overlay is flipped(black line moves to the right side of the overlay) seems like the main culprit.

Do you mean this: 
![image](https://user-images.githubusercontent.com/17603532/141062199-a5bd9dd6-be97-4e47-ae94-9b2a57494162.png)

The flag moves to the left side of the line when the cursor is far to the right. But this works without flickering on my machine.

otbutz:
I thought that it had something to do with the orientation of the thin black line(left or right of the moving detail flag). But i think that was just a coincidence.

Small reproducible case:

https://graphhopper.com/maps2/?point=48.983014%2C12.090756&point=49.019437%2C12.104966&profile=bike&layer=OpenStreetMap

Moving my cursor in the area of 370m height between 1.50km and 2.50km doesn't render any flag for me on Firefox/Chrome.

easbar:
I can reproduce this as well. It should be fixed by #155.

otbutz:
I manually changed the `<svg class="heightgraph-container">` elements CSS in Firefox to `pointer-events: auto;` and can confirm that it works.

easbar:
Fixed by #155

PR #155, opened 2021-11-10
No more flickering when hovering the path details diagram

easbar:
This fixes the flickering of the path detail window described in #134. 
We set `pointer-events: all` in this commit: https://github.com/graphhopper/graphhopper-maps/pull/75/commits/924ee0574d97e82aa4b8d56afe8688e65564c1c0 of this PR: https://github.com/graphhopper/graphhopper-maps/pull/75. Not sure if there is a particular reason we used `all` so far, but using `auto` fixes the issue for me.

### DIFF
--- a/src/pathDetails/PathDetails.module.css
+++ b/src/pathDetails/PathDetails.module.css
@@ -7,7 +7,7 @@
 
 .innerDiv {
     position: relative;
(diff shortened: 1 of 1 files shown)


karussell:
Thanks - works for me too!

Issue #143, opened 2021-10-11
Sidebar height needs change


karussell:
Probably this was introduced for #13.

1. the default height is too large - too much white space

![image](https://user-images.githubusercontent.com/129644/136796454-022937ef-93d1-4609-b4d9-b2055b135e9f.png)

2. the last suggestions only show up with scrolling

![image](https://user-images.githubusercontent.com/129644/136796597-c0106419-6bff-4fba-a195-5c3be0b06238.png)

3. when too many via points exist the result will be hidden:

![image](https://user-images.githubusercontent.com/129644/136797124-ca7b5c9a-0e99-4b64-aed4-01ca6933b7c5.png)

For the last point I would prefer to have a scroll bar only for the via points not the entire sidebar. Or maybe a similar "collapsing" solution like we use for mobile?

UnrealValentin:
> when too many via points exist the result will be hidden

Shouldn't the expected behavior be that it would expand till it reaches the bottom end of the screen before a scrollbar appears? At least that is the behavior in gmaps, I think

PR #164, opened 2021-11-22
143 search with fullsize autocomplete

Janekdererste:
Started to work on another version of the search implementation. At the moment:
- A textbox can be selected
- Once the user starts typing the fullsize Autocomplete opens
- When the fullsize autocomplete is closed (back button, esc, remove all text), or something was selected (tab, enter, valid coordinate) the last input is selected so that the user can tab through the interface
- The Current location selection is a button next to the input at the moment because the fullsize autocomplete only opens if text was entered. 

I am currently not happy with the current location selection since there are two tiny buttons next to each other next to the input. Especially on mobile, it is hard to target the correct button. 

I broke stuff on mobile, because I wanted to get things right for the large screen layout first. 

@karussell and maybe @easbar can you please try it and give feedback on how this attempt feels?


easbar:
> I am currently not happy with the current location selection since there are two tiny buttons next to each other next to the input. Especially on mobile, it is hard to target the correct button.

Yes, I don't really like this either.

To me it seems a bit strange that once we start typing the sidebar disappears. Can it not 'just' extend beyond the sidebar (in vertical direction) and offer a scrollbar if there is not enough room to show all the search results? 

I also looked at the version where the search results are shown to the right of the sidebar (#166), but this does not feel so intuitive to me. But let's see what @karussell has to say.

Janekdererste:
>To me it seems a bit strange that the sidebar disappears once we start typing. Can it not 'just' extend beyond the sidebar (in vertical direction) and offer a scrollbar if there is not enough room to show all the search results?

Hm, I don't exactly know what you mean I think. Do you have something in mind like it is at the moment but not within the scroll container? I guess I could try around a little and move the autocomplete box from #166 around a little. With this portal the autocomplete box can actually go anywhere.

easbar:
I mean currently we have this:

image

So the search results are only partly visible, even though there is enough space to show them. I thought the search results could just 'leave' the side bar container, or the side bar could just grow with the search results panel?

For many via points there won't be enough space below the side bar:

image

But this might be ok since we have the scrollbar.

karussell:
I like it :)

The current location is indeed tricky. It could be added e.g. as 6th entry of the search result or something.

I would prefer this over #166 which feels a bit unusual but maybe one has to get used to it.

> I thought the search results could just 'leave' the side bar container, or the side bar could just grow with the search results panel?

How would this work when you do the auto complete at the bottom of the screen (when you have many via points)? The suggestions will be outside of the screen and you cannot scroll to make the suggestions visible. We discussed that one could show the suggestions in that case above the input but we decided against it for now, probably it will feel similar unusual as #166.

Another option would be to have different behaviour depending on where you type. I.e. for input boxes at the top the suggestion panel will be outside of the sidebar container and for input boxes on bottom of the screen the panel will be inside the sidebar to have the scrollbar. But for many via points this will be probably too complex as the scrollbar might make it possible to move one input from top to bottom, i.e. behaviour of one input box has to change depending on where it is on the screen.

easbar:
> How would this work when you do the auto complete at the bottom of the screen (when you have many via points)? The suggestions will be outside of the screen and you cannot scroll to make the suggestions visible.

But currently it actually is possible to scroll in this case:

image

You also suggested this: 

> For the last point I would prefer to have a scroll bar only for the via points not the entire sidebar. Or maybe a similar "collapsing" solution like we use for mobile?

This also sounds like a good solution?

Regarding the present PR: Could we show the fullsize autocomplete on top of the side bar, just like we do now, but without hiding the sidebar? Anyway, I don't have a strong opinion here. If you guys can agree on something just go ahead :)

easbar:
Btw can we update graphhopper.com/maps2 to latest master? The first point in #143:

> 1. the default height is too large - too much white space

seems to be fixed already. 

And can we update this branch to master as well, so we can see the preview on aws? The build currently fails so it is not deployed.

easbar:
And one more thing: I would not mind if the font used for the search results was a bit smaller. At least the second line could use a smaller font and the padding could also be a bit reduced:

image

This way we might run into the problem that there is not enough space for the search results a bit less often and there will also be more space to show longer street names etc.

Janekdererste:
>The build currently fails so it is not deployed.

I accidentially branched off your package.lock branch. I will merge the master branch.

Janekdererste:
> Thanks! I like this :)

Alright, I will keep working on this approach then. I will put the current location into the autocomplete list for now. One has to start typing to see it though. Maybe I'll have a better idea. 

> just like we do now, but without hiding the sidebar? 

I don't like the complete collapse of the side bar either. I will try with adding empty search results, so that the sidebar keeps its height while search results are loaded.

karussell:
I see the following options:

1) This PR. We can solve the current location issue similar to mobile, ie. show it for 0 characters and maybe also for 1 and 2. But I also understand that this option is distracting as the view always jumps to the top of the screen when you type. Also "mouse-only" users will have to heavily move the mouse, i.e. click "add point" and then click on the first suggestion, click "add point" etc

2) One interesting implementation I see is when the suggestion panel would be shown **above** the input for inputs at the bottom of the screen. I initially found this strange but probably this is the best solution, because the strange "above input" situation will only happen for many via-points and only for inputs at the bottom. IMO this is better compared to a jump of the view for **all** inputs (except the first) like in this PR.

3) In this regard the PR #166 is also interesting, but IMO it feels a bit strange for all inputs and additionally collides with path details and the map too much.

4) A (too?) simple solution could be to use master but reduce the height of the suggestion panel a bit (like suggested from @easbar) **and** add some more "useful" things to the sidebar below the inputs (bigger horizontal profile list? some basic options? custom model editor?), so that the current whitespace isn't that obvious and the scrollbar never shows up for a few via points.

> > How would this work when you do the auto complete at the bottom of the screen (when you have many via points)? The suggestions will be outside of the screen and you cannot scroll to make the suggestions visible.

> But currently it actually is possible to scroll in this case

Yes, but it won't be possible when it is outside of the sidebar.

> Regarding the present PR: Could we show the fullsize autocomplete on top of the side bar, just like we do now, but without hiding the sidebar? 

What advantages would this have?

> >  For the last point I would prefer to have a scroll bar only for the via points not the entire sidebar. Or maybe a similar "collapsing" solution like we use for mobile?

> This also sounds like a good solution?

This solution would have the downside of not being able to get a fast overview of the points and the order and also every time you start typing or one of the input gets the focus the point list has to collapse and the view jumps (I think).

> And one more thing: I would not mind if the font used for the search results was a bit smaller. At least the second line could use a smaller font and the padding could also be a bit reduced:

:+1: 

> Btw can we update graphhopper.com/maps2 to latest master? The first point in #143:

Hm, strange it should be automatically done.

easbar:
> Yes, but it won't be possible when it is outside of the sidebar.

Ok, maybe not automatically. But in principle it should be possible to not limit the vertical extent of the search results by the sidebar and show a scrollbar nonetheless?

> What advantages would this have?

I just thought it felt odd that the sidebar disappeared just because I started typing, especially for the first input boxes where there is sufficient space anyway.

> One interesting implementation I see is when the suggestion panel would be shown above the input but only for inputs at the bottom of the screen. I initially found this strange, because the strange "above input" situation will only happen for many via-points and only for inputs at the bottom. IMO this is still a bit better compared to a jump of the view for all inputs (except the first) like in this PR.

Yes, maybe doing something special only for the lower input boxes is the way to go. I also would not mind showing search results above the (lower) input boxes, I think.

> Hm, strange it should be automatically done.

But it is definitely an outdated version as for example #153 and #155 also do not seem to be included.

easbar:
> I just thought it felt odd that the sidebar disappeared just because I started typing, especially for the first input boxes where there is sufficient space anyway.

But actually also for the lower input boxes. For example when I start typing at the last input box the input field moves all the way to the top of the screen. I think I would be interested in a version of #166 where the search results are shown below the input box for the upper boxes and above for the lower ones. But not sure how much effort is needed for this.

Janekdererste:
Alright, to get an impression what option 1. could look like, I added initial empty items to the result list, so that the sidebar doesn't collapse enirely. Also, I put the current location option as last item into the autocomplete component. This can of course be further tweaked.

> I would be interested in a version of #166 where the search results are shown below the input box for the upper boxes and above for the lower ones.

I will create an example of this. I guess having this as a pop up is the most intuitive way on Desktop.

Janekdererste:
Okay, please have a look at #169 . I used the portals approach for this one. The autocomplete now basically works as before but the autocomplete items are mounted in another container so that they can overflow our scrolling sidebar. In case there is not enough space on the bottom the items are shown above. 

This would be my favorite.

Janekdererste:
We are going to use https://github.com/graphhopper/graphhopper-maps/pull/169 for this. Therefore closing this PR

PR #165, opened 2021-11-22
Update more dependencies

### LINKED ISSUE #148: update dependencies where possible
(and maybe we should all upgrade to a more recent npm version to avoid these package-lock.json problems?)

easbar:
Fixes #148. There are still some deprecation warnings left, but at least I don't see vulnerabilities anymore. The only thing I was not able to upgrade was jest (26->27). The problem I was running into is that our `flushPromises` helper uses `setImmediate` which is no longer part of the `jsdom` `testEnvironment` we use.


easbar:
Not sure why but according to the changed line count in `package-lock.json` this PR seems to reduce the number of (sub-)dependencies.

Issue #162, opened 2021-11-16
Map zooms to both the info bbox and the route bbox if url contains points


easbar:
Loading a url like: https://graphhopper.com/maps2/?point=51.311911%2C10.205488&point=51.332152%2C10.260636&profile=car&layer=MapTiler

triggers a request to the `/info` endpoint as well as a request to the `/route` endpoint. These requests result in `InfoReceived` and `RouteRequestSuccess` actions, for both of which we update the map bounds in `ViewPortStorte#reduce`. This means we first fit the map to the info bounds and shortly after fit it to the route bounds (or vice versa). Apparently the `/info` request is always (?) faster so we first fit the map to the info bounds and eventually zoom to the route bounds as it should be. However, this is rather coincidental. Using OpenLayers (#159) I also ran into the opposite case where we first fit the map to the route bounds, but then end up with the map being zoomed out entirely (fit to the info bounds). Not sure why this never seems to happen in current master, but I was able to reproduce this by adding a small delay for the `InfoReceived` event.

This could even be related to longer map loading times, because first we 'zoom out' (triggering map tile requests) only to 'zoom in' again just after that.

Btw we had the same issue in #82, but I think this was only partly fixed/hidden so far.

easbar:
I think the two requests are triggered here:

https://github.com/graphhopper/graphhopper-maps/blob/8331e8ebafd4c01d5bdee889ccdbf5a04d876a64/src/index.tsx#L57-L62

Can we somehow first parse the url so in case some points are given (and a route request will be fired) we can prevent the update of the map bbox that is based on the /info response?

easbar:
For now we disabled zooming to the info bbox entirely as it is always the entire planet except for local setups, so I'll close here. We can consider zooming to the info box in #117 again.

Issue #159, opened 2021-11-12
Migrate from Mapbox to OpenLayers


easbar:
Still not decided whether or not we should do this, but if yes we should probably do it soon. 

Some pros and cons compared to Mapbox (just considering the issues we ran into so far)

**Pro**:

* faster raster tile rendering
* no blurry raster tiles (#156) when `constrainResolution` is enabled
* very active development compared to Mapbox, which is only developed under a proprietary license now
* was able to fix the long press issue on iOS
* (minor) commented out layers seem to work out of the box (#57)

**Con**:

* slower vector tile rendering (important for /spt, not so much for /mvt)
* if we consider vector tiles to be too slow with OL we cannot take advantage of some of the corresponding benefits, like custom map styling. Still we could use the Mapbox GL JS integration of OL.
* cannot pitch/tilt the map -> try PerspectiveMap extension

I started to explore this migration here: https://github.com/graphhopper/graphhopper-maps/tree/ol_exp 
The live instance is running here: http://gh-maps-react.s3-website.eu-central-1.amazonaws.com/ol_exp/
See also the discussion here: https://github.com/graphhopper/graphhopper-maps/issues/132

This is the current list of todos for the ol_exp branch:

* [ ] fix remaining svg marker issues, see commit message: f2213e24e457ca90b56ce2e5ae01370abbbf116c, but at least the cursor appearance is now fixed
* [x] make markers draggable
* [x] add context menu (including the long touch handler)
* [x] put back path details layer
* [x] add map attributions
* [x] allow click selection of the paths
* [ ] restrict map zoom/pan so we cannot pan around the globe
* [ ] put back view store test
* [x] use same mouse cursors for map as in master
* [ ] bug/issue: when refreshing map with a route the map flickers and sometimes even ends up being zoomed out all the way to world view -> this is caused by a bug in master (#162)
* [ ] test (also on mobile)

easbar:
[...]

Issue #117, opened 2021-06-22
Draw map bounding box


easbar:
Only really needed when a smaller map than planet is used, but still useful IMO. We should also change the camera position such that the map extract is shown if it's not the entire planet.

Issue #167, opened 2021-11-22
NavBar#parseUrlAndReplaceQuery seems a bit broken


easbar:
Take [this route](https://graphhopper.com/maps2/?point=52.610119%2C13.319295&point=52.560485%2C13.233999&point=52.50787%2C13.256825&point=52.477149%2C13.380564&point=52.524685%2C13.459852&point=52.601365%2C13.391376&point=52.565597%2C13.349329&profile=car&layer=OpenStreetMap) for example. Move a marker and then click the back button. Multiple `/route` requests are fired and we can also notice that the camera is moving in a weird way and there are rapid updates of the heightgraph (both because of the many route requests).

I noticed this when I was trying to find a solution for #162. Let's take a look at the code:

https://github.com/graphhopper/graphhopper-maps/blob/fbe18e16a4ab0b3e3d9451246acf3ab13a832528/src/NavBar.ts#L89-L113

This code runs when we hit the 'back' button of the browser, because we add an event listener for the `popstate` event in the `NavBar` constructor. At first sight `this.isIgnoreQueryStoreUpdates = true` looks like we can manipulate the state without triggering any routing requests. But this flag only blocks the code that is run in `onQueryStateChanged`, which updates the url. The standard receivers of the query state change are still registered and will fire the routing requests. Because the request was valid before, it will remain valid when we remove each of the points and there will be a routing request after every point we remove. Similarly, there will be even more routing requests when add points in the next line, and even another one when we set the profile.

`parseUrlAndReplaceQuery` is not only called when the `popstate` event occurs, but also when the app loads for the first time (see `index.tsx`). In this case only one route request is fired, but this is rather coincidental it seems. When we refresh the page the `QueryStore` will be in it's initial state, namely with two (empty) points and an **empty** profile name:

https://github.com/graphhopper/graphhopper-maps/blob/fbe18e16a4ab0b3e3d9451246acf3ab13a832528/src/stores/QueryStore.ts#L69-L84

This means the query store state is invalid initially and this check in `routeIfAllPointsSet`:

https://github.com/graphhopper/graphhopper-maps/blob/fbe18e16a4ab0b3e3d9451246acf3ab13a832528/src/stores/QueryStore.ts#L200-L204

[...]

PR #168, opened 2021-11-22
Clean up initial map load, disable zoom animation

easbar:
Instead of actually fixing #162, which seems not so trivial, I thought we could just disable the 'zoom to info bbox' feature entirely for now. The info bbox is always equal to the entire planet view anyway, except for local development with smaller maps. We could re-consider this for example in #117, but right now the initial page load seems a lot more important, especially because we also still need to decide which map library we should continue working with (#159).

Therefore, I disabled the viewport update that is triggered by the `/info` endpoint here. I also disabled the viewport transition animation to reduce the initial map flickering as much as possible. Reloading a url that includes points still does not work perfecty though, because for raster tiles we still see a very quick view transition from the world view to the route bbox. To get rid of this maybe we even have to delay the map loading until we received the route response(?).


easbar:
> Reloading a url that includes points still does not work perfecty though, because for raster tiles we still see a very quick view transition from the world view to the route bbox. To get rid of this maybe we even have to delay the map loading until we received the route response(?).

Maybe we could just first parse the url and then choose the bbox given by the points as initial map view. Even without actually calculating the route first.

karussell:
> To get rid of this maybe we even have to delay the map loading until we received the route response(?)

Yes, I think this is the only way to avoid this when points are specified. (For the old GH maps the world wide view still loads but shows up only very quickly and so it is not really disruptive.)

easbar:
> (For the old GH maps the world wide view still loads but shows up only very quickly and so it is not really disruptive.)

For this PR it is kind of the same, but unfortunately we see the world view a bit longer compared to GH maps 1 and it is therefore more distracting. Strangely, I see a blurred 'Deutschland' label when I use a route in Germany, as if some intermediate tiles between the world view and the route bounding box were loaded. At least this happens for a longer route like this:

http://localhost:3000/?point=52.933932%2C9.936341&point=48.786663%2C12.418884&profile=bike&layer=OpenStreetMap

But for a shorter one like this:

http://localhost:3000/?point=49.131117%2C12.296425&point=48.786663%2C12.418884&profile=bike&layer=OpenStreetMap

it seems fine and I do not even get to see the world view. For the `ol_exp` branch (OpenLayers) it is similar. I also briefly see the world view and the blurry 'Deutschland' label. 

>  Yes, I think this is the only way to avoid this when points are specified.

I think you are right. First I thought we could just estimate the bounding box from the url points before we send the route query. Unfortunately it is hard to get a good estimation for this for all cases. The tricky thing is that if we do not get a good estimate for the bounding box from the url points and then update the map view again when the routing request returns we are at risk of loading even more tiles than we would load if we did not try to estimate the bounding box at all. This is because the higher level tiles are larger than e.g. the world view tiles, because they show far more details. 

I tried delaying the map rendering until the route request returns using Mapbox here: 
https://github.com/graphhopper/graphhopper-maps/commit/a9c51b28bc2c201e7077e0c81edcd9f0584c1fc3 
and using OpenLayers here:
https://github.com/graphhopper/graphhopper-maps/commit/22ac0c36cef0da68da8877d8748aa00514992eb1
With these changes I no longer see any map flickering in both cases. 

You can compare this here:
http://gh-maps-react.s3-website.eu-central-1.amazonaws.com/master/?point=51.73643%2C9.512061&point=50.722208%2C10.517404&profile=bike&layer=OpenStreetMap
http://gh-maps-react.s3-website.eu-central-1.amazonaws.com/load_map_after_route/?point=51.73643%2C9.512061&point=50.722208%2C10.517404&profile=bike&layer=OpenStreetMap
http://gh-maps-react.s3-website.eu-central-1.amazonaws.com/ol_exp_render_map_after_route/?point=51.73643%2C9.512061&point=50.722208%2C10.517404&profile=bike&layer=OpenStreetMap

Compared to `master` there is no longer an animation from a larger to the route bbox view for and `load_map_after_route`. We only load the tiles we actually need to show the route bbox. `ol_exp_render_map_after_route` does the same using OpenLayers.

Note that my implementations are more of a hack to try this out. For example we won't see a map at all if the route request fails, e.g. if one of the points cannot be found. We would have to implement this more thoroughly and probably this is also related to #167. I think we need the following logic:

1) Load /info. If there is no response show the world map with a clear error (the app won't work when there is no routing backend)
2) load /route if there are url points and delay the map rendering until we receive the route bbox. If there are either no url points or if the route request fails show the /info bbox.

But unfortunately, even with this approach there is a problem: When we add the custom model editor we have to deal with requests that take a lot longer than what we are used to right now. This might mean that delaying the map rendering until the initial route request returns will no longer be an option. Or we need to show some kind of loading animation for this case and maybe show the map for a preliminary bounding box that is based on the url points for such slow requests.

karussell:
Thanks for this!

> But unfortunately, even with this approach there is a problem: 

Ah, indeed this is problematic and also the implementation seems to be rather tricky.

> and maybe show the map for a preliminary bounding box that is based on the url points for such slow requests.

Yes, sounds good. This would also (often) reduce the required animation time and might reduce flickering even when enabled for fast returning routes? (And for the most routes the initially loaded map tiles won't change due to the route geometry.)

Another possibility is to store the current bbox in the URL. And if none exists in the URL we load the bbox from /info. And if we calculate a route we attach the bbox to the URL. This way one could share the URL with an arbitrary bbox (could be seen as an advantage :) ).

easbar:
> This would also (often) reduce the required animation time and might reduce flickering even when enabled for fast returning routes? (And for the most routes the initially loaded map tiles won't change due to the route geometry.)

I'm not sure. Like I said, first I thought it is a good idea to use the url points as a first estimation, but when I tried this I found that even when the final bounding box only changed a bit there was still some kind of flickering effect and many more tiles were loaded. I just meant for slow routing request we might have no other option if we don't want to show the map at all until the route was calculated. But this would make the implementation a lot more complicated indeed.

> Another possibility is to store the current bbox in the URL. And if none exists in the URL we load the bbox from /info. And if we calculate a route and auto zoom to it we attach this bbox to the URL or update the old. This way one could share the URL with an arbitrary bbox (could be seen as an advantage :) ).

Yes, this could be a very good solution indeed. Would you keep the url in sync with the camera position all the time? Qwant Maps does this too. They store the zoom level and center point, which might be the easiest option, but I also wonder how this works when sharing urls between devices with different screen sizes:

https://www.qwant.com/maps/#map=13.24/47.2111331/7.2227028

And yes, this way it would be possible to share a location even without calculating a route. For GH maps 1 we were able to do this with a single marker: https://graphhopper.com/maps/?point=51.404036%2C13.521234. For maps 2 this does not work as well, because the map is zoomed out to the world view: https://graphhopper.com/maps2/?point=51.404036%2C13.521234

easbar:
> Like I said, first I thought it is a good idea to use the url points as a first estimation, but when I tried this I found that even when the final bounding box only changed a bit there was still some kind of flickering effect and many more tiles were loaded. 

I just pushed this here: https://github.com/graphhopper/graphhopper-maps/commit/5aff626bd1d532f7ff5894a1f6716f7bc167e003 and you can try it here:

http://gh-maps-react.s3-website.eu-central-1.amazonaws.com/clean_up_initial_map_load_estimate_bbox/?point=51.73643%2C9.512061&point=50.722208%2C10.517404&profile=bike&layer=OpenStreetMap

Actually, now it does not look so bad to me anymore?!

easbar:
... and to complete this little collection here is the version where we estimate the bbox from the url points using OpenLayers: https://github.com/graphhopper/graphhopper-maps/commit/844d7d5af410ba2956ee194be1ee905abd029996

http://gh-maps-react.s3-website.eu-central-1.amazonaws.com/ol_exp_estimate_bbox/?point=51.73643%2C9.512061&point=50.722208%2C10.517404&profile=bike&layer=OpenStreetMap

karussell:
> but I also wonder how this works when sharing urls between devices with different screen sizes

The center and zoom should solve this (?) and would be probably better than my proposed bbox.

One question then is when we store a new history state (for back button vs. up-to-date URL).

> I just pushed this here: 5aff626 and you can try it here

Nice :+1: 

But indeed it loads further tiles and cancels the older ones. Wouldn't have this expected for most cases.

btw: Unfortunately it does not fix the problem for the mapbox version that when you open the URL in a background tab it does not update to the bbox of the route compared to the situation when you open it in a foreground tab. (the OL version does not have this problem)

easbar:
> The center and zoom should solve this (?) and would be probably better than my proposed bbox.

Not necessarily, because the visible map area depends not only on the center and zoom, but also on the map size (in pixels). So it could happen that the zoom level that is calculated on a larger screen is too large (too far zoomed in) to show the entire route on a smaller screen.

> One question then is when we store a new history state (for back button vs. up-to-date URL).

Yes, good question, but maybe we can just not create a history state when we zoom/pan the map, but do create a history state whenever we calculate a new route?

> But indeed it loads further tiles and cancels the older ones. Wouldn't have this expected for most cases.

As long as the older ones are canceled it should be ok (?). Can you notice a difference in performance for all these approaches and also Mapbox vs. OpenLayers?

> Unfortunately it does not fix the problem for the Mapbox version that when you open the URL in a background tab it does not update to the bbox of the route compared to the situation when you open it in a foreground tab

When I right-click and choose 'Open Link in New Tab' it works, but not when I use the middle mouse button of my laptop:

http://gh-maps-react.s3-website.eu-central-1.amazonaws.com/clean_up_initial_map_load_estimate_bbox/?point=51.73643%2C9.512061&point=50.722208%2C10.517404&profile=bike&layer=OpenStreetMap

No idea why, but this could be something unrelated? I also don't know what the difference between using the middle mouse button and choosing 'Open Link in New Tab' is. The tab is opened in the 'background' in both cases. And yes, for the OL version it works for both Mapbox and OL.

This 'estimate bbox from url points' approach seems like the easiest thing we can do right now and at least it is an improvement over the current master? But yes, it is still not ideal. The next best thing is probably storing the camera state in the URL.

karussell:
> Not necessarily, because the visible map area depends not only on the center and zoom, but also on the map size (in pixels). So it could happen that the zoom level that is calculated on a larger screen is too large (too far zoomed in) to show the entire route on a smaller screen.

Ah, ok. Now I understand what you mean. Yes, indeed. So the bbox is better in this regard as we can say that the specified bbox should be taken as the minimum bbox.

> Yes, good question, but maybe we can just not create a history state when we zoom/pan the map, but do create a history state whenever we calculate a new route?

Yes, still we would have to update the URL as otherwise the shared URL is something different. Or we have a separate "share" button like before. (probably unavoidable due to the custom model anyway)

> As long as the older ones are canceled it should be ok (?)

It is a bit an overhead we could avoid but yes this should be fine.

> Can you notice a difference in performance for all these approaches and also Mapbox vs. OpenLayers?

between the mapbox versions: no
between mapbox and openlayers: openlayers seems to be a bit faster (and mapbox is still ugly for rasters ;) )
Could it be that for mapbox the routes are sequentially drawn after the map is drawn? And openlayers draws the routes as fast as it gets them and then somehow is able to plot the map later. So OL feels more responsive. (comparing here "estimate_bbox" branches)

> No idea why, but this could be something unrelated?

It also occurs in master and is unrelated but I thought that it was related to how master handles the bbox and hoped it was fixed here.

> This 'estimate bbox from url points' approach seems like the easiest thing we can do right now and at least it is an improvement over the current master? But yes, it is still not ideal. The next best thing is probably storing the camera state in the URL.

:+1:

easbar:
> Ah, ok. Now I understand what you mean. Yes, indeed. So the bbox is better in this regard as we can say that the specified bbox should be taken as the minimum bbox.

Yes, better in this regard, but also a bit more complicated, i.e. we need to calculate the bbox all the time using the center point, zoom level and current screen width/height. But that should be doable.

> Yes, still we would have to update the URL as otherwise the shared URL is something different

Yes, but this has always been like this. When we calculate a route in GH maps 1, and then change the camera position and share the url, our updated camera position won't be included.

> Or we have a separate "share" button like before. (probably unavoidable due to the custom model anyway)

Yes, probably. This is related to: #25 

> between mapbox and openlayers: openlayers seems to be a bit faster

Ok, my main motivation for this PR was to get rid of any overhead, like the additional zoom to the info bbox and the zoom animation so we can rule out these factors to be the cause of the slow map loading. If OpenLayers is still faster with these changes that is an argument against Mapbox.

>  (and mapbox is still ugly for rasters ;) )

Do the rasters look ugly only in your Browser or also on your phone? 

> Could it be that for mapbox the routes are sequentially drawn after the map is drawn? And openlayers draws the routes as fast as it gets them and then somehow is able to plot the map later. So OL feels more responsive. (comparing here "estimate_bbox" branches)

I don't know why the route rendering is so slow.

karussell:
> Do the rasters look ugly only in your Browser or also on your phone?

Yes, I think the reason is that mapbox allows the partial zoom

> Actually, now it does not look so bad to me anymore?!

Feel free to merge this as this is certainly an improvement over master.

easbar:
> Yes, I think the reason is that mapbox allows the partial zoom

I thought on my iPhone it's not as bad as on my Laptop. Maybe because of different screen resolutions and maybe this means we could improve this also for Desktop still. But the issue with the partial zooming will remain anyway (I think).

> Feel free to merge this as this is certainly an improvement over master.

I opened https://github.com/graphhopper/graphhopper-maps/pull/171 now. Will leave this here as is for later reference. Closing for now.

PR #169, opened 2021-11-23
143 auto complete just as before

### LINKED ISSUE #143: Sidebar height needs change

Janekdererste:
Yet another attempt on autocomplete. This is going to fix #143 

Our Sidebar has some requirements. It should:
- Display an unlimited number of QueryPoints
- Be navigable by keyboard
- Show autocomplete suggestions when users type stuff into the search boxes
- Display a variable number of routing results
- Display expanded routing results which have a variable number of items themselves

This requires the entire sidebar to be scrollable. When introducing this feature the autocomplete Popup broke because it was using absolute positioning for floating above the sidebar. With the sidebar being scrollable, this approach is no longer possible because the scroll container "swallows" the popup and the user would have to scroll to see results. To have both, a scrollable container and a popup, We now have a transparent popup container in the large-screen layout into which we mount a popup container with a Portal. The caveat of this solution is a small break in incapsulation of components. The popup needs to know the id of the scroll container and of the popup container to mount itself into the correct position and to monitor scroll events to update its position.


easbar:
This is also my favorite, very nice!

easbar:
One (probably easy to fix) thing I noticed: If I start typing in one input box the search results popup opens, but if I click another input box and start typing there another popup opens and the first one does not close.

karussell:
Yes, I too think this makes the most sense - thank you! 

Although initially it is a bit strange to have the suggestions above the input I think this is still rare (you need to have many via points) and those "power users" will get used to this :)

Janekdererste:
Very good. I'll proceed with this approach then. 

One thing I noticed: We select the first option by default on TAB. I find this not very intuitive, since we put the current_location item into the list if it is empty yet. This way one can't TAB through all the input boxes quickly. I would keep things the way they are but not select the current location on tab.

easbar:
I also like how you can either press 'up' to enter the list from the bottom or 'down' to enter it from the top :+1:

easbar:
Very nice! :tada:

Issue #132, opened 2021-09-29
Initial page load is still slow compared to original GH maps


easbar:
@boldtrn suggested the following:

 - [x] enable brotli/gzip
 - [ ] measure page load using: 
  - https://developers.google.com/speed/pagespeed/insights/?url=https%3A%2F%2Fgraphhopper.com%2Fmaps%2F&tab=desktop
  - https://developers.google.com/speed/pagespeed/insights/?url=https%3A%2F%2Fgraphhopper.com%2Fmaps2%2F&tab=desktop
  or lighthouse
 - [x] bundle size -> not much difference to old GH Maps
 - [ ] why are two icons loaded: /assets/favicon-16x16.png and /assets/apple-touch-icon-1024x1024.png
 - [x] ~~initially maptiler does 3 requests to fonts (3*43kb) - can we reduce to 1 request?~~ > we use raster tiles by default since #159
 - [x] ~~load map with raster tiles in background first: https://support.maptiler.com/i818-load-the-map-visually-more-quickly-using-a-raster-background but currently they are blurred!? -> https://support.maptiler.com/i851-incorrect-usage-of-512px-tiles~~ -> > we use OpenLayers with raster tiles by default since #159
 - [ ] profile in chrome
 - [x] try on latest FF94 as there were some [perf improvements](https://bugzilla.mozilla.org/show_bug.cgi?id=1732049) for a specific Mapbox GL JS case -> we use OpenLayers not Mapbox since #159
 - [x] ~~try without wrapper~~ -> we use OpenLayers without the react-map-gl wrapper since #159
 - [x] try raster -> we use OpenLayers with raster tiles by default since #159
 - [x] ~~try different vector tiles provider~~ -> we use raster tiles by default since #159

and I think we already figured that maptiler vector tiles are slow compared to mapbox, even on their website. However, this should not affect the page load when we use raster tiles?

boldtrn:
As a side-note, if you look at the library size of leaflet vs. maplibre, it is significantly larger (contains all the code to render vector tiles :)).

https://bundlephobia.com/package/maplibre-gl@2.0.0-pre.5
https://bundlephobia.com/package/leaflet@1.7.1

easbar:
Maybe it is worth looking at these experiments I did a while back again: https://github.com/easbar/gh_maps_experiments

karussell:
[...]

PR #171, opened 2021-12-08
Estimate initial bbox from url points, disable zoom animation

easbar:
Following the discussion in #168 here I disabled the bbox update after the `/info` request and disabled the zoom animation. For URLs with points I estimate the bounding box of the resulting `/route` request to prevent loading the world view just before we zoom to the route. This should improve the map/page loading performance because less tiles need to be loaded (#132).


src/stores/ViewportStore.ts
```diff
@@ -35,21 +34,24 @@ export default class ViewportStore extends Store<ViewportStoreState> {
...
 
-    constructor(routeStore: RouteStore, isSmallScreenQuery: () => boolean) {
+    constructor(routeStore: RouteStore, initialBBox: Bbox | null, isSmallScreenQuery: () => boolean) {
         super()
         this.routeStore = routeStore
         this.isSmallScreenQuery = isSmallScreenQuery
+        if (initialBBox) {
+            const updatedState = calculateLatLngFromBbox(this.state, initialBBox, isSmallScreenQuery())
+            super.setState(updatedState)
+        }
```
easbar, line 44: @Janekdererste I wasn't sure how to do this, because `getInitialState` is called in the super-constructor and subclasses like `ViewportStore` have no chance to make use of the constructor parameters before `getInitialState` is called. Here I'd like to pass the `initialBBox` via the constructor and actually make use of it to set the initial state.

src/stores/ViewportStore.ts
```diff
@@ -35,21 +34,24 @@ export default class ViewportStore extends Store<ViewportStoreState> {
...
         return {
-            // todo: initial values do not really matter, because we immediately update the state?!..
-            width: 800,
-            height: 600,
-            longitude: 11,
-            latitude: 48,
-            zoom: 5,
-        }
+            width: window.screen.width,
+            height: window.screen.height,
```
easbar, line 50: Using the `window.screen.height/width` here is useful to get a rough estimation of the bounding box size, but maybe it could be more precise. The windows size is not equal to the map size and I think this is the reason why we still see a small flickering of the map: First the map is zoomed such that the markers are right at the map borders and then zoomed out a bit more. @Janekdererste do you have an idea how we could improve this?

src/stores/Store.ts
```diff
@@ -34,7 +34,7 @@ export default abstract class Store<TState> implements ActionReceiver, NotifySta
 
     abstract reduce(state: TState, action: Action): TState
 
-    private setState(value: TState) {
+    public setState(value: TState) {
```
Janekdererste, line 37: I don't think we want to have this public.

src/stores/ViewportStore.ts
Janekdererste: Since the only way to alter a store's state should be through actions, I guess you need an action for this. Like UpdateBboxAction.

Initialize all the stores in ```index.ts``` then parse the bbox and issue an UpdateBboxAction to the Dispatcher.

src/index.tsx
```diff
@@ -74,3 +76,19 @@ function getApiKey() {
     // use graphhopper api key from url or try using one from the config
     return url.searchParams.has('key') ? url.searchParams.get('key') : config.keys.graphhopper
 }
+
+function parseURLBBox() : Bbox | null {
```
Janekdererste, line 80: Ultimately this could be in the ```NavBar``` component as well. Once you update the viewport state by dispatching an action, you could put this logic there.

src/stores/ViewportStore.ts
Janekdererste: I think this is a reasonable heuristic, since the map is pretty much the same size as the window. This gets called before the first render pass and before any css is applied though. Therefore, it is not really possible to guess the map size correctly

Janekdererste (approved):
I would think that some of this will become obsolete once the viewport information is stored in the map again. I'd assume that this is the case once you merge your PR with the switch to openlayers.

src/stores/ViewportStore.ts
easbar: Ok, I did this in my last commit.

src/stores/Store.ts
easbar: Ok yes it's no longer needed since I use an action to set the initial state now.

Issue #136, opened 2021-10-01
Improve visual appearance of context menu and elevation popup


otbutz:
Old:

![grafik](https://user-images.githubusercontent.com/22315436/135609193-3d8bdfe0-68bd-4e75-be1d-5d11a3e1c175.png)

New:

![grafik](https://user-images.githubusercontent.com/22315436/135609265-937d3d7f-c43e-4232-9f92-0f0d9044f8ec.png)

* properly align
* add images
* separator for additional actions

easbar:
The elevation popup is also ugly:

![image](https://user-images.githubusercontent.com/17603532/141061137-42dea1f2-a04e-4d58-87b0-eb9ef568dbf5.png)

Before it looked like this: 

![image](https://user-images.githubusercontent.com/17603532/141061328-9450dfe6-bb88-4ae3-b298-8b875aec22b1.png)

It used to be an `<svg>` layer on top of the map box, but I was not sure how to do this with Mapbox, hence we just use the default popup currently.

easbar:
After the OpenLayers migration the popups look like this now: 

![image](https://user-images.githubusercontent.com/17603532/151128787-3c7e2127-4f24-46a6-b8d2-3911b50b99f7.png)

![image](https://user-images.githubusercontent.com/17603532/151128835-5612023a-a3bf-45c0-98dd-3c75a08815af.png)

We still need to improve this. See comments [here](https://github.com/graphhopper/graphhopper-maps/blob/92f788355e5028216a98a2bd960b03ee534e60ce/src/layers/ContextMenu.module.css#L1-L4), [here](https://github.com/graphhopper/graphhopper-maps/blob/92f788355e5028216a98a2bd960b03ee534e60ce/src/layers/PathDetailPopup.module.css#L1-L4) and [here](https://github.com/graphhopper/graphhopper-maps/blob/92f788355e5028216a98a2bd960b03ee534e60ce/src/layers/PathDetailPopup.tsx#L36-L37)

otbutz:
I think we can close this one. The current state looks pretty good.

easbar:
I kind of liked this old elevation 'flag', but yeah there are probably more important things to work on :)

easbar:
I'd rather keep it open still, there are also some related todos in the code. But I'll change the title and remove the context menu part.

Issue #135, opened 2021-10-01
A left click in the map should close the context menu


otbutz:
![grafik](https://user-images.githubusercontent.com/22315436/135608898-df04f2b9-2b7c-493c-a430-5783b3ef139a.png)

easbar:
This is/was a bug in react-map-gl: https://github.com/visgl/react-map-gl/issues/1389.
#158 will 'silently' update react-map-gl from 5.3.16 to 5.3.17 which should fix the issue. Unfortunately it does not it seems...

easbar:
But this is probably a problem in our code. I confirmed that 5.3.17 fixes the problem upstream using their "controls" demo: https://github.com/visgl/react-map-gl/tree/master/examples/controls). They use this:

```typescript
            closeOnClick={true}
            onClose={setPopupInfo}
```

and we do not use `onClose` yet. If I use this the context menu closes immediately after it was opens. Maybe this is even related to the context menu not appearing on mobile (https://github.com/graphhopper/graphhopper-maps/pull/149)

easbar:
... or maybe not: I can reproduce another problem upstream: https://github.com/visgl/react-map-gl/issues/1389#issuecomment-965315354

easbar:
This issue is still not fixed after #159. It's actually worse now: The context menu not only stays open when we click the map, but also when we pan the map. I'm actually undecided whether or not it should close when we move the map, but it should definitely close when we click the map.

easbar:
Yes, this works perfectly, thanks! I tried FF + Chrome. The popup stays open when we move the map, but since it closes when we click I think this is fine (or maybe even better).

easbar:
Sorry, I should have tried don mobile before merging. I cannot open the context menu anymore. Tried on iOS 14.8 using Safari,FF&Chrome

easbar:
Openlayers implemented a long-press touch event for their drawing feature: https://github.com/openlayers/openlayers/issues/10904#issuecomment-693241627
Maybe this helps figuring out how we can open (and close) the context menu.

Janekdererste:
Ah, very good. I’ll look into that.

easbar:
> I'm actually undecided whether or not it should close when we move the map, but it should definitely close when we click the map.

Now I think it should definitely stay open when we move the map, because quite often we have to move the map a bit after opening the menu to see all the menu entries.

PR #176, opened 2022-01-18
Use OpenLayers instead of Mapbox

easbar:
This is a follow-up of #159, this time trying to actually do this migration rather than exploring its implications.

Here is a small walk-through:

1) package.json/dependencies: I removed the react-map-gl dependency (the react/mapbox wrapper including mapbox) obviously and added OpenLayers 6 (`ol`) instead. I also added `ol-mapbox-style` to show the MapTiler tiles, even though we might remove this entirely since we rather focus on raster tiles for now. I also had to add the `geojson` dependency directly, because we were using it here and there and we got it from Mapbox so far.

2) tsconfig.json: I just modified this to make the app run. Possibly we can revert parts of this especially if we remove `ol-mapbox-style`.

3) Now to the more interesting changes: How do we create and show the map? First thing we need is the OL `Map` object. I'm not entirely sure where we should create it but right now I simply create it in `map.ts` and there is also a reference to it so we can import it from everywhere. The next thing we need to do is attach the map to the DOM. For this I added `MapComponent.tsx` which uses `useRef` to create a container div for the map. It simply receives the map object as a prop. 

4) The next thing we need to do is add some layers to the map. Let's start with the background layer. If we just used the same layer forever we could just add it to the map initially and would be done with it. But since we can change the layer via the layers option menu we need code that first removes the existing layer and then adds the new layer. This should run whenever the `styleOption` changes, however we don't want to run this code for every render call of our `App` component. Therefore I use `useEffect` and put this `useEffect` into another function called `useBackgroundLayer` (a custom hook: https://reactjs.org/docs/hooks-custom.html). I added the other layers (like query points and paths) the same way. One advantage here is that we can add layers without passing all the props we need for the different layers into our previous `MapComponent`. `useBackgroundLayer`, just like `MapComponent`, simply receives the map (and the `styleOption`) as a function parameter in `App.tsx`. Note that if we did not use `useEffect` our 'remove-old-and-add-new-layer'-code would run for every render call.

5) We also need to adjust the map view for certain actions like `RouteRequestSuccess` or `SetSelectedPath` etc. Therefore I use an `ActionReceiver` that is registered at our `Dispatcher`. It is called `MapActionReceiver` and created in `index.ts` similar  to our stores. It also gets a reference to the map objects so it can adjust the map view for whatever `Action` we want. 

6) This is basically it, most of the remaining code is simply about adding the other layers. I'll add a todo list and a few more comments inline.


src/NavBar.ts
```diff
@@ -90,6 +90,11 @@ export default class NavBar {
         //const parseResult = NavBar.parseUrl(this.appContext.location.href)
         const parseResult = this.parseUrl(window.location.href)
 
+        // estimate map bounds from url points if there are any. this way we prevent loading tiles for the world view
+        // only to zoom to the route shortly after
+        const bbox = this.getBBoxFromUrlPoints(parseResult.points.map(p => p.coordinate))
+        if (bbox) Dispatcher.dispatch(new SetInitialBBox(bbox))
+
```
easbar, line 97: This is basically what we do in #171

src/api/graphhopper.d.ts
```diff
@@ -54,6 +54,7 @@ export interface RoutingProfile {
 }
 
 export interface Path extends BasePath {
+    // todo: do we really need the geojson dependency? it was included in mapbox but its not included in OL
```
easbar, line 57: do we?

src/layers/ContextMenu.module.css
```diff
@@ -0,0 +1,37 @@
+/* todo: this is just copy-pasta from https://openlayers.org/en/latest/examples/popup.html, I have no idea what all this means,
+   and the popup looks ugly still, at the very minimum the background color needs to be set though
+   the popup design needs an overhaul anyway... https://github.com/graphhopper/graphhopper-maps/issues/136
+   */
+.popup {
```
easbar, line 5: Compared to the react-map-gl wrapper OpenLayers does not provide a popup component. Therefore I used a simple div and copy-pasted some css from the OpenLayers examples for the context menu and path details popup. I'm pretty sure this can be improved...

src/layers/PathDetailPopup.module.css
```diff
@@ -0,0 +1,37 @@
+/* todo: this is just copy-pasta from https://openlayers.org/en/latest/examples/popup.html, I have no idea what all this means,
+   and the popup looks ugly still, at the very minimum the background color needs to be set though
+   the popup design needs an overhaul anyway... https://github.com/graphhopper/graphhopper-maps/issues/136
+   */
```
easbar, line 4: Same as for the context menu

src/layers/PathDetailPopup.tsx
```diff
@@ -0,0 +1,48 @@
...
+    useEffect(() => {
+        const position = pathDetails.pathDetailsPoint
+            ? fromLonLat([pathDetails.pathDetailsPoint.point.lng, pathDetails.pathDetailsPoint.point.lat])
+            : undefined
+        overlay?.setPosition(position)
+    }, [pathDetails.pathDetailsPoint])
+
+    return (
+        // todo: use createMapMarker from heightgraph?
+        // {createMapMarker(point.elevation, point.description)}
```
easbar, line 37: This todo is just copied from master

src/layers/createMarkerSVG.ts
```diff
@@ -0,0 +1,34 @@
...
+    color: string
+    number?: number
+    size?: number
+}
+
+// todo: this is mostly duplicated from `Marker.tsx`. To use `Marker.tsx` we would probably need to add ol.Overlays, i.e.
+//       create a div for each marker and insert the svg from `Marker.tsx`.
+export function createSvg({ color, number, size = 0 }: MarkerProps) {
+    return `<svg aria-hidden="true" focusable="false" role="img" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 384 512" width="${
+        // todo: we do not use width in Marker.tsx, but without this the markers are not shown in Firefox :( (but they are shown in Chrome...)
```
easbar, line 15: * I had to duplicate the svg from `Marker.tsx`, because I needed it as a string to use it as Icon src for the style image of the query points vector layer. The alternative could be using an ol.Overlay for each marker, but then we need to somehow create a div (using useRef?) for each overlay. Considering the issues I ran into here (see below) this might be the better option?

* The marker positions are off. Apparently the `translate` style property doesn't work correctly

* The marker numbers/text is only shown in Chrome. In Firefox it is missing or so small that it is invisible.

* I had to change the colors that are given as hex strings to 'rgb(red, green, blue)'. Otherwise the markers were invisible?!

src/map/Map.module.css
```diff
@@ -1,3 +1,8 @@
 .contextMenu {
     z-index: 3;
 }
+
+.mapContainer {
+    height: 100%;
+    width: 100%;
+}
```
easbar, line 8: Not sure, but I needed this to make the map visible

src/stores/MapOptionsStore.ts
```diff
@@ -96,6 +98,7 @@ const tfCycle: RasterStyle = {
     attribution:
         osmAttribution +
         ', <a href="https://www.thunderforest.com/maps/opencyclemap/" target="_blank">Thunderforest Cycle</a>',
+    tilePixelRatio: 2,
```
easbar, line 101: This way the raster tiles aren't blurry like they were for Mapbox

src/stores/MapOptionsStore.ts
```diff
@@ -193,10 +199,9 @@ const styleOptions: StyleOption[] = [
     kurviger,
     mapillion,
     osmDe,
-    // The original client has these but those options yield cors errors with mapbox yields a cors error
-    // lyrk,
-    // wanderreitkarte,
-    // This works but is extremely slow with mapbox
+    lyrk,
+    wanderreitkarte,
+    // This one is extremely slow with mapbox and openlayers?!
```
easbar, line 204: lyrk and wanderreitkarte work with OpenLayers while they did not for Mapbox. sorbian is still very slow

src/stores/MapOptionsStore.ts
```diff
@@ -208,7 +213,7 @@ export default class MapOptionsStore extends Store<MapOptionsStoreState> {
                 `Could not find tile layer specified in config: '${config.defaultTiles}', using default instead`
             )
         return {
-            selectedStyle: selectedStyle ? selectedStyle : mapTiler,
+            selectedStyle: selectedStyle ? selectedStyle : osmOrg,
```
easbar, line 216: We probably want to change this to Omniscale

tsconfig.json
```diff
@@ -1,5 +1,6 @@
 {
     "compilerOptions": {
+        "allowJs": true,
```
easbar, line 3: If this important we can try to get rid of this again but I needed it to make the code compile

src/map/map.ts
```diff
@@ -0,0 +1,38 @@
...
+import { fromLonLat } from 'ol/proj'
+import { MapIsLoaded } from '@/actions/Actions'
+import { defaults as defaultControls } from 'ol/control'
+
+let map: Map | undefined
+
+export function createMap(): Map {
+    map = new Map({
+        view: new View({
+            multiWorld: false,
```
easbar, line 12: We still need to restrict map zoom/pan so we cannot pan around the globe?! (currently we can)

src/stores/MapActionReceiver.ts
```diff
@@ -0,0 +1,56 @@
...
+    private readonly isSmallScreenQuery: () => boolean
+
+    constructor(map: Map, routeStore: RouteStore, isSmallScreenQuery: () => boolean) {
+        this.map = map
+        this.routeStore = routeStore
+        this.isSmallScreenQuery = isSmallScreenQuery
+    }
+
+    receive(action: Action) {
+        // todo: port old ViewportStore.test.ts or otherwise test this
```
easbar, line 26: I'll take a look at this, but leaving this as a reminder here for now.

src/layers/createMarkerSVG.ts
Janekdererste: It would be nice to have this mounted as a ```Marker.tsx``` component. This would indeed mean that we would have a wrapper div for each point in a route. One would then create a new rendering context for each point. Maybe this would be too complicated. 

The way this works now, is that ol renders the svg on the canvas as an image [directly](https://github.com/openlayers/openlayers/blob/30d27a1d59113617cc0247dd862cdc763ca76216/src/ol/render/canvas.js#L460). I'd assume that this is the most efficient way to do it. 

Leave it like this, fix the translation and the number?

src/map/map.ts
Janekdererste: I don't have an oppinion on that

src/stores/MapActionReceiver.ts
```diff
@@ -0,0 +1,56 @@
+import { Action, ActionReceiver } from '@/stores/Dispatcher'
```
Janekdererste, line 1: This looks quite nice so far. I think this is a good direction.

tsconfig.json
Janekdererste: Interesting. Which parts would not compile?

Janekdererste (commented):
This looks very good overall. The layers part being inside custom hooks turned out pretty I think. It is quite easy to get an idea what's going on. 

I noticed, that the context menu sometimes doesn't open, after I have set some points via the context menu. It might be related to the change between wide and narrow layouts. Also, the context menu doesn't close anymore at all, except one selects something from it. I think this is a separate issue, which we have somewhere already thoug.

src/map/map.ts
karussell: When we pan around the globe: will the longitudes get an offset of +360°? For leaflet this was the case and our API has issues with that so we limited the paning for leaflet.

karussell: (will try this when I try out this PR)

src/api/graphhopper.d.ts
easbar: Ok I was more thinking about not using it at all. But let's just keep it for now.

src/layers/ContextMenu.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+export default function ContextMenu({ map, route, queryPoints }: ContextMenuProps) {
+    const [menuCoordinate, setMenuCoordinate] = useState<Coordinate | null>(null)
+    const [overlay, setOverlay] = useState<Overlay | undefined>()
+
+    const container = useRef()
+
+    useEffect(() => {
+        const overlay = new Overlay({
+            element: container.current as any,
```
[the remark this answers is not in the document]
easbar: Yes. The only thing we need is `container.current!` because the `Overlay` expects a non-null value.

src/layers/PathDetailPopup.module.css
easbar: Ok. And we have this issue already: https://github.com/graphhopper/graphhopper-maps/issues/136

src/layers/UsePathDetailsLayer.tsx
```diff
@@ -0,0 +1,70 @@
...
+import VectorSource from 'ol/source/Vector'
+import { Stroke, Style } from 'ol/style'
+import { GeoJSON } from 'ol/format'
+import { fromLonLat } from 'ol/proj'
+
+const highlightedPathSegmentLayerKey = 'highlightedPathSegmentLayer'
+
+export default function usePathDetailsLayer(map: Map, pathDetails: PathDetailsStoreState) {
+    useEffect(() => {
+        removePathSegmentsLayer(map)
```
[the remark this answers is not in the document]
easbar: Hm the layer is related to the path details diagram so `usePathDetailsLayer` kind of make sense. But what it actually does is just highlight segments of the path that are above/below a certain elevation. I don't know.. I'm not sure if renaming is an improvement here. I'll add a comment that says what this layer does for now.

src/layers/createMarkerSVG.ts
easbar: Ok that makes sense. Can you have a go at this?

src/map/Map.module.css
easbar: Yes, you're right this also works. I changed it.

src/map/map.ts
easbar: No, I panned around the world a few times and the longitude of the requests seems to always be in the range -180, 180

tsconfig.json
easbar: Hm, I just tried again and it seems to work also without this now :)

easbar:
> This looks very good overall. The layers part being inside custom hooks turned out pretty I think. It is quite easy to get an idea what's going on. 

Ok cool, thanks for your review.

> I noticed, that the context menu sometimes doesn't open, after I have set some points via the context menu. It might be related to the change between wide and narrow layouts. 

Hm, can you reproduce this consistently somehow?

> Also, the context menu doesn't close anymore at all, except one selects something from it. I think this is a separate issue, which we have somewhere already thoug.

Yes, I think it master it does not close when we click (see #135). But it does close when we pan the map. In this branch it only closes when we select an action. Let's fix both in #135?

easbar:
> Yes, I think it master it does not close when we click (see #135). But it does close when we pan the map. In this branch it only closes when we select an action. Let's fix both in #135?

I just tried this briefly and it is easy to close the menu on click. Just add

```typescript
        map.on('click', () => {
            overlay?.setPosition(undefined)
            setMenuCoordinate(null)
        })
```

in `ContextMenu.tsx`. The tricky thing is to make this not interfere with the long press handler, make it work on mobile etc, so let's do this later in #135. I'm also wondering: Should we close it when we pan the map or not?

src/stores/MapOptionsStore.ts
easbar: And shall we even keep map tiler?

karussell: Let's use Omniscale as default (but keep map tiler)

src/map/map.ts
karussell: Yes, perfect :+1:

src/stores/MapOptionsStore.ts
karussell: But as maptiler and mapilion are so slow we should probably mark them as experimental (e.g. gray out the label a bit?)

easbar: Ok I made Omniscale the default now. If the two vector layers are so slow that they should not be used why not remove them altogether? Or is there any case where they would be useful?

karussell:
Thanks a lot for this @easbar, great work!

Everything works and due to the raster tile speed and precision the project feels even more polished :)

I played a bit around and tried to find the longest shortest path ;)

The marker offset is not yet perfect. E.g. have a look about the start of this route which seems to be in the sea instead of the land: http://localhost:3000/?point=-34.356099%2C18.497489&point=66.638129%2C179.986909&profile=car&layer=OpenStreetMap (Interestingly the marker in current master is also wrong but in a different way)

Then there seems to be some bbox issue (but very likely this is related to #53) when the route cannot be found:
http://localhost:3000/?point=66.641111%2C-179.99621&point=66.64092%2C179.99635&profile=car&layer=OpenStreetMap

> Also, the context menu doesn't close anymore at all, except one selects something from it. I think this is a separate issue, which we have somewhere already thoug.

Yes. In master it closes when moving the map but now it still stays open.

easbar:
Thanks for testing!

> The marker offset is not yet perfect. 

Yes, see also this discussion: https://github.com/graphhopper/graphhopper-maps/pull/176#discussion_r787073295

> Then there seems to be some bbox issue 

Yes, this is strange. I did not figure out what causes this yet. First we estimate the bbox from the url points (as in #171) and then we adjust it to the actual route once it is calculated. But if no route could be calculated the bbox should just remain the same. But for some reason this does not work...

src/map/map.ts
easbar: We can limit the panning like this: 

```typescript
    const sw = fromLonLat([-180, -75])
    const ne = fromLonLat([180, 85])
    map = new Map({
        view: new View({
            extent: [sw[0], sw[1], ne[0], ne[1]],
```

but I don't really have a preference here

src/stores/MapOptionsStore.ts
karussell: For testing and improving I would keep them. We could then ask / contract e.g. OL maintainers to improve this somehow :)

src/stores/MapOptionsStore.ts
karussell: Isn't the avoidance of fractional numbers for zoom what the blurry avoids? `constrainResolution: true` See your comment https://github.com/openlayers/openlayers/issues/12963#issuecomment-961834559

And tilePixelRatio (additionally) improves blurry for high dpi devices? (but for mapbox we had these "`2x`" in the URL)

(then also the question would be: could we set it to 1 for low dpi devices somehow?)

easbar:
> Then there seems to be some bbox issue

I fixed this here: dd0ade875155781fb57bf6d643e347706654519e. Now we see the map also for urls for which no route can be found. The problem was that at the time we set the initial bounding box the map's screen size was still undefined as it hadn't been rendered at that point yet. Therefore we need to estimate the map size as well.

src/stores/MapOptionsStore.ts
easbar: > Isn't the avoidance of fractional numbers for zoom that avoids the blurry look? (constrainResolution: true)
See your comment openlayers/openlayers#12963 (comment)

Yes, avoiding fractional zoom levels is part of it. But `tilePixelRatio` is still needed I think. As far as I know we need to tell OpenLayers also about the resolution of our tiles. From the documentation: https://openlayers.org/en/latest/apidoc/module-ol_source_XYZ-XYZ.html

> The pixel ratio used by the tile service. For example, if the tile service advertizes 256px by 256px tiles but actually sends 512px by 512px images (for retina/hidpi devices) then tilePixelRatio should be set to 2.

So we need to set this number depending on the tile service. 

> but for mapbox we had these "2x" in the URL

Yes, for some tile providers adding `2x` changes the resolution of the tiles they send, but either way we need to tell OpenLayers the resolution of the tiles we feed to it?

> (then also the question would be: could we set it to 1 for low dpi devices somehow?)

Do you see blurry tiles on some device, or what do you want to improve here?

easbar: Also see this and the following comments: https://github.com/graphhopper/graphhopper-maps/issues/159#issuecomment-973917832

src/map/Map.module.css
easbar: Hm, this works in Firefox, Chrome and also when using the mobile view in dev tools. But for the aws deployment the map stayed invisible on mobile for @karussell and me (even though it worked for desktop). So I reverted it to my initial version.

Janekdererste: Hm, this is odd.

easbar: Yes, but I tried it twice and we both saw the same results. Feel free to try it again, but otherwise I'll leave as it is this for now.

easbar:
I'll add separate issues for some of the things we discussed here.

PR #181, opened 2022-01-31
detect retina to avoid requesting unnecessary high res images

karussell:
also add retina for omniscale too (it was a bit blurry on my phone)

(it seems intellij ignores the default formatting -> fixed)


easbar:
Nice this looks better now also on my phone :+1:

PR #183, opened 2022-02-14
Add custom model editor to support custom routing

### LINKED ISSUE #17: custom model box

easbar:
This PR adds support for routing using so-called 'custom models' and fixes #17. A custom model is a set of rules written in JSON language that affects the outcome of the route calculation. Custom models are not used standalone, but rather they are modifications of existing profiles. So for example one can take the car profile and modify the routing using a custom model to reduce the speed on motorways. See the documentation for more details [here](https://docs.graphhopper.com/#section/Custom-Model).

Users shall be able to enter such custom models using a text field (the 'custom model editor'). For this and to guide the user when writing custom models we developed an interactive text field that supports syntax checking, error-highlighting and auto-completion [here](https://github.com/graphhopper/custom-model-editor). This shall be added here.

Routing with custom models comes at a cost: The response will be slower, sometimes much slower. Therefore, we clearly need to distinguish two cases:

1) Routing using the built-in profiles (car, foot, bike etc.) as we do so far
2) Routing with custom models (car, foot, bike etc. + custom model).

For 2) there is a catch: Custom models are not supported for all built-in profiles. Currently we offer `car,foot,bike,hike,racingbike,scooter,small_truck,truck,mtb` on http://graphhopper.com/maps, but custom models are only available for `car,small_truck,foot,bike,scooter` (not for `truck,hike,mtb,racingbike`). This might be different for self-deployed GraphHopper instances etc. Unfortunately, currently we do not have this information available on the client side, so we might just ignore this for now as we do for the current GH maps app.

Here I added a simple checkbox to the sidebar that opens/closes the custom model box right in the sidebar, just like we do for current GH maps. This is probably subject to change, because the space in the sidebar is rather limited (horizontally, but also vertically when there are instructions, many via-points etc.). Also we might need even more space for the following features that we might add at a later time:

1) Custom models can also be used to modify the routing in a certain area. Add support to visually select an area on the map.
2) Certain custom model rules are very common, like 'avoid motorways'. Add a combo box to insert such rules into the text field to make using custom models easier.
3) Custom model rules are applied based on the attributes of the different roads. It would be useful to visualize these attributes, e.g. 'show all roads with "road_class == primary"` etc.

So maybe at some point we need extra layout changes for these things, add tabs to the sidebar etc.

Opening/closing the custom model editor using the checkbox also defines whether we are in 'standard' or 'custom model' mode. This way we can easily compare routes with and without a custom model by opening and closing the text field.

For an open custom model box we need to distinguish three further cases:

1) A custom model was entered but it is not even valid JSON (or the text field is just empty). We cannot parse the model at all and it is therefore not possible to perform a route request.
2) Valid JSON was entered for the custom model and we can parse from it a JavaScript object, but it's not a valid custom model. The user needs to fix the custom model first. We allow sending the routing request regardless using the keyboard shortcut Ctrl+Enter, but otherwise the user needs to fix the custom model first.
3) A valid custom model was entered and we are ready to send the routing request and show the route.

The custom model editor/text field helps us dealing with these three cases. For example when there are syntax errors they will already be highlighted in red and the user can hover the corresponding text to see more details about the error. The same applies when the syntax is correct, but the custom model is invalid. Still, we need a way to alert the user about these errors. For example if the user adds a new query point we usually send a routing request right away, but if the custom model is invalid we should rather tell the user about this. I use the same error notification (`ErrorAction`) we use to report error responses (like 'point not found' etc.) for this.


src/sidebar/CustomModelBox.tsx
```diff
@@ -0,0 +1,84 @@
+import 'codemirror/lib/codemirror.css'
+import 'codemirror/addon/hint/show-hint.css'
+import 'codemirror/addon/lint/lint.css'
+// todonow: this belongs to this app and we should not take it from the demo...
+import 'custom-model-editor/demo/style.css'
```
easbar, line 5: I was not sure how to add (where to put the file) global css rules as we need them for codemirror like `.cm-s-default`.

src/stores/QueryStore.ts
```diff
@@ -26,6 +29,9 @@ export interface QueryStoreState {
     readonly currentRequest: CurrentRequest
     readonly maxAlternativeRoutes: number
     readonly routingProfile: RoutingProfile
+    readonly customModelEnabled: boolean
+    readonly customModelValid: boolean
+    readonly customModel: CustomModel | null
```
easbar, line 34: I added three new fields to the app's state: `customModelEnabled` simply tells us whether we are in 'custom model mode' or not (and right now it also determines whether the custom model editor is shown or not). `customModelValid` tells us whether or not the model is valid as it is determined by the editor component. `customModel` is the current custom model. We cannot just set the `customModel` field to null in case it is invalid, because we like to be able to send requests also for invalid models. Note that `customModel` is **not** used as backend for the editor's actual content. This is because if we managed the editor's content using react we would need to manage all of the editor's state (like cursor position, annotations etc.) with react (or at least keep it up to date somehow).  Right now I don't see a real advantage of this.

src/stores/QueryStore.ts
```diff
@@ -140,7 +156,7 @@ export default class QueryStore extends Store<QueryStoreState> {
                 queryPoints: newPoints,
             }
 
-            return this.routeIfAllPointsSet(newState)
+            return this.routeIfReady(newState)
```
easbar, line 159: I renamed this method because the old name was no longer accurate (because of custom model baut also because of the routing profile before).

src/sidebar/CustomModelBox.tsx
```diff
@@ -0,0 +1,84 @@
...
+            else {
+                try {
+                    const customModel = JSON.parse(instance.value)
+                    Dispatcher.dispatch(new SetCustomModel(customModel, false))
+                } catch (e) {
+                    Dispatcher.dispatch(new SetCustomModel(null, false))
+                }
+            }
+        }
+        Dispatcher.dispatch(new SetCustomModel(initialCustomModel, true))
```
easbar, line 43: We need to update the app state here. This is the price we pay for not using the app's custom model state as the editor's backend. Afterwards we just update the `customModel` state whenever the editor evaluated the input's validity (see `validListener` above).

easbar:
Currently it looks like this:

image

but I just tried to make it work first, so I am sure there is room for improvements

easbar:
> I tried the branch and it seems to work. Let's fix the Dispatch of the Error Actions, then this could be merged I think.

Thanks for taking a look!

> One thing to think about is optics. Would you like the editor to be where it is at the moment, or should we think of something else? Instead of a check box we could also have an icon for this and a "closing x". Also, 15rem is quite large, if the model is simple.

You are more than welcome to improve on this. I just put it anywhere really so I could go on implementing the functionality.

src/stores/ErrorStore.ts
```diff
@@ -21,11 +21,7 @@ export default class ErrorStore extends Store<ErrorStoreState> {
                 lastError: action.message,
                 isDismissed: false,
             }
-        } else if (
-            action instanceof DismissLastError ||
-            action instanceof RouteRequestSuccess ||
-            (action instanceof SetCustomModelBoxEnabled && !action.enabled)
-        ) {
+        } else if (action instanceof DismissLastError || action instanceof RouteRequestSuccess) {
```
easbar, line 24: Ah now I see what you meant. Maybe the `instanceof RouteRequestSucess` check does not really belong here either?

src/sidebar/CustomModelBox.tsx
```diff
@@ -69,6 +60,15 @@ export default function CustomModelBox({ enabled, encodedValues }: CustomModelBo
...
 
+    const triggerRouting = useCallback(
+        (event: React.KeyboardEvent<HTMLInputElement>) => {
+            if (event.ctrlKey && event.key === 'Enter') {
+                dispatchCustomModel(editor.value, isValid, true)
+            }
+        },
+        [editor, isValid]
+    )
+
```
easbar, line 71: Not sure if this is better or worse, but we can also use the event handling of the text component using this API: https://github.com/graphhopper/custom-model-editor/blob/693a44097271ed84f603f46efbfcfc56ab97b608/src/custom_model_editor.js#L84-L86

This is how we did it in GH maps 1.0 as well: https://github.com/graphhopper/graphhopper/blob/876ffaa9fb990e2ea812c8956091809084025dcd/web-bundle/src/main/resources/com/graphhopper/maps/js/main-template.js#L153

And is it possible to send the request/json without validation when pressing Ctrl+Enter? @karussell should we do this? This is what we did for the old maps client.

easbar:
Maybe it would be useful to add some kind of hint that advises the user to use the Ctrl+Space shortcut for auto-completion?

src/sidebar/CustomModelBox.tsx
karussell: > @karussell should we do this?

:+1:

easbar: Using Ctrl+Enter we can now force the request regardless of the client side validation. https://github.com/graphhopper/graphhopper-maps/pull/183/commits/78c612cfea1c13ed0e319e9f64d70a885de88955

easbar:
@Janekdererste Are you planning to work on this? Or is there something I should change? Or can we merge this and do further improvements later? 

You wrote:

> Let's fix the Dispatch of the Error Actions, then this could be merged I think.

But unfortunately I don't know what you meant?

Here are a few things that could be still done, but that might not be strictly required before merging:

* [ ] Better layout: Do not squeeze the custom model box into the sidebar, also see my initial post. 
* [ ] The 'custom' checkbox used to enable/disable the custom model is quite ugly as you already pointed out. Same for the 'Help' and 'Example' buttons underneath the custom model box. I just added them as they were in GH maps 1. The 'Example' button also seems a bit useless to me, because the example custom model is used already initially.
* [ ] Better UX: We do not automatically trigger a routing request when we edit the custom model. This is intended, because there would just be too many requests if we did this for every edit. On the other hand this might be confusing for users, because it is hard to understand what they need to do to trigger the request after editing the model. In GH maps 1 there is a 'Search' button which is greyed out whenever the mode becomes invalid. I think this makes it quite obvious one has to click this button after editing the model. Here we decided against adding such a search button, so we need to focus one of the query point boxes and press 'Enter' to  trigger the request instead. The other option is pressing Ctrl+Enter, but this is even less obvious.

Issue #179, opened 2022-01-26
Fix Marker offset, via-point labels in Firefox and maybe re-use Marker.tsx


easbar:
After #159 we have the following issues related to the query point markers:

* [ ] The query point markers are slightly shifted from the coordinate they should represent. See this discussion: https://github.com/graphhopper/graphhopper-maps/pull/176#discussion_r787073295_ 
* [ ] The via-point labels are no longer visible in Firefox, see here: https://github.com/graphhopper/graphhopper-maps/blob/8518a15177877fe2a940a72845a6faf855c27fa4/src/layers/createMarkerSVG.ts#L24
* [ ] We duplicate the svg that is already in `Marker.tsx`: https://github.com/graphhopper/graphhopper-maps/blob/8518a15177877fe2a940a72845a6faf855c27fa4/src/layers/createMarkerSVG.ts#L11-L12

easbar:
> We duplicate the svg that is already in Marker.tsx: 

So far our svg markers are simply drawn on the map canvas, but with OpenLayer's overlays we can create separate div elements where we can put our marker component into. I tried this here: f44c46f9ea52c21fb357160df1155b07b5a246e5
I managed to show the markers using this approach and also the via-point labels are shown correctly in Firefox this way.
Unfortunately, I ran into a problem where it seems like both OpenLayers and React are interfering with the DOM causing React to crash: 

https://github.com/graphhopper/graphhopper-maps/blob/f44c46f9ea52c21fb357160df1155b07b5a246e5/src/layers/QueryPoints.tsx#L15-L18

The other issue is that we can no longer use the mouse wheel to zoom the map when the cursor is above one of the map markers. This is because React handles the mouse events and does not forward them to OpenLayers. We will probably run into a similar issue when we try to implement the dragging interaction with this Overlay approach.

Since I found a way to fix the marker offsets without overlays (#184) maybe it's not worth it to switch to the Overlay approach, but then we also need to find another fix for the via-point label issue.

Janekdererste:
For the crash I would have to look into it to get an idea. 

> The other issue is that we can no longer use the mouse wheel to zoom the map when the cursor is above one of the map markers.

I would hope that this can be solved with a little bit of css by setting ```touch-action: none```

karussell:
> firefox: adding 'px' helps

Don't ask me why ;)

PR #184, opened 2022-02-15
Fix marker offsets

easbar:
Some work for #179. I adjusted the marker offsets so the tip of the marker symbols actually points to the coordinates they represent.

### DIFF
--- a/src/layers/createMarkerSVG.ts
+++ b/src/layers/createMarkerSVG.ts
@@ -14,9 +14,8 @@ export function createSvg({ color, number, size = 0 }: MarkerProps) {
     return `<svg aria-hidden="true" focusable="false" role="img" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 384 512" width="${
         // todo: we do not use width in Marker.tsx, but without this the markers are not shown in Firefox :( (but they are shown in Chrome...)
         size
-    }" height="${size}" style="cursor: pointer; stroke: none; fill: ${hexToRgb(color)}; transform: \`translate(${
-        -size / 2
-    }px, ${-size}px)"><path d="${MARKER_PATH}"/>${
+    }" height="${size}" style="cursor: pointer; stroke: none; fill: ${hexToRgb(color)};">
+    <path d="${MARKER_PATH}"/>${
         number === undefined
             ? '<path d="' + INNER_CIRCLE + '" fill="white" />'
             : '<circle cx="192" cy="190" r="120" fill="white" />'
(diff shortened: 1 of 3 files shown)


src/map/Marker.tsx
```diff
@@ -28,7 +28,7 @@ export function MarkerComponent({ color, number, size = 0 }: MarkerProps) {
             xmlns="http://www.w3.org/2000/svg"
             viewBox="0 0 384 512"
             height={size}
-            style={{ ...markerStyle, fill: color, transform: `translate(${-size / 2}px, ${-size}px` }}
```
easbar, line 31: This was broken in multiple ways: First of all, the closing parenthesis of the `translate` function was missing so this `transform` did not do anything at all. Second, if anything we only want to translate in the vertical direction (not horizontally). Third, we want to translate vertically only by half of the size of the symbol (the origin is at the center without translation, but the tip of the marker is at the bottom). Finally, size is `0` here currently, so the translation won't do nothing even after fixing the aforementioned issues. Why is size 0? And why do we still see the markers (I mean in the sidebar. For the markers in the map we do not use `Marker.tsx` currently (see #179))?

src/layers/createMarkerSVG.ts
```diff
@@ -14,9 +14,8 @@ export function createSvg({ color, number, size = 0 }: MarkerProps) {
     return `<svg aria-hidden="true" focusable="false" role="img" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 384 512" width="${
         // todo: we do not use width in Marker.tsx, but without this the markers are not shown in Firefox :( (but they are shown in Chrome...)
         size
-    }" height="${size}" style="cursor: pointer; stroke: none; fill: ${hexToRgb(color)}; transform: \`translate(${
-        -size / 2
-    }px, ${-size}px)"><path d="${MARKER_PATH}"/>${
```
easbar, line 19: This `transform: translate` function was also broken (see my comment in `Marker.tsx`). It had the closing parenthesis, but there was an extra backtick before `translate`. It seems like we can use `translate` to move the markers, but then we won't see the entire marker any longer because it is moved out of the svg's `viewBox`. I therefore used the `displacement` option for the `Icon` in `UseQueryPointsLayer` instead.

src/map/Marker.tsx
Janekdererste: It could be that we set the width of the element from the svg's container. Though the element's style should be applied last 🤔

easbar: Ok, but for now / as long as we do not use `Marker.tsx` for the map markers I think we can just remove this transform/translate because apparently it did not do anything.

easbar:
I'll merge here as this fixes the problem and we can still think about using another approach later.

PR #186, opened 2022-03-29
167 fix parseurl query

### LINKED ISSUE #167: NavBar#parseUrlAndReplaceQuery seems a bit broken

Janekdererste:
* Add action which sets all routing params at once
* Change navbar so that it sets the query points and the routing profile at once
* make the query store digest the setrutingparamsatonce action

fixes #167

### DIFF
--- a/src/NavBar.ts
+++ b/src/NavBar.ts
@@ -1,7 +1,7 @@
 import { coordinateToText } from '@/Converters'
 import { Bbox, RoutingProfile } from '@/api/graphhopper'
 import Dispatcher from '@/stores/Dispatcher'
-import { AddPoint, RemovePoint, SelectMapStyle, SetInitialBBox, SetVehicleProfile } from '@/actions/Actions'
+import { SelectMapStyle, SetInitialBBox, SetRoutingParametersAtOnce } from '@/actions/Actions'
 // import the window like this so that it can be mocked during testing
 import { window } from '@/Window'
 import QueryStore, { Coordinate, QueryPoint, QueryPointType, QueryStoreState } from '@/stores/QueryStore'
@@ -61,7 +61,7 @@ export default class NavBar {
                     coordinate: coordinate,
                     isInitialized: true,
                     id: i,
-                    queryText: '',
+                    queryText: coordinateToText(coordinate),
                     color: '',
                     type: QueryPointType.Via,
                 }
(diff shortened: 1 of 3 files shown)


Janekdererste:
Hups, just noticing that I gave the wrong ticket number in the branch 😅

src/stores/QueryStore.ts
```diff
@@ -84,14 +85,7 @@ export default class QueryStore extends Store<QueryStoreState> {
     }
 
     reduce(state: QueryStoreState, action: Action): QueryStoreState {
-        if (action instanceof SetPoint) {
-            const newState: QueryStoreState = {
-                ...state,
-                queryPoints: QueryStore.replacePoint(state.queryPoints, action.point),
```
easbar, line 90: Very minor, but it looks like you changed the order of the if/else blocks for `SetPoint` and `InvalidatePoint` here for no real reason? Without this change it would be a bit easier to see what changed in this PR.

easbar (approved):
Thanks! I tested this PR and I can confirm that the rapid route updates after clicking the back button I described in #167 are fixed.

src/NavBar.ts
```diff
@@ -95,26 +95,28 @@ export default class NavBar {
...
-        // assuming that at least two points should be present add un-initialized points if necessary
-        for (let i = this.queryStore.state.queryPoints.length; i < 2; i++) {
-            Dispatcher.dispatch(new AddPoint(i, { lat: 0, lng: 0 }, false))
-        }
-
-        // add routing profile
-        Dispatcher.dispatch(new SetVehicleProfile(parseResult.profile))
+        // we want either all the points replaced from the url or we just have one and we want to replace the one default
+        // one
+        const points = parseResult.points.length > 2 ? parseResult.points : this.fillPoints(parseResult.points)
```
easbar, line 100: When do we have only a single point and with which default one do we want to replace this?

easbar (approved):

src/NavBar.ts
Janekdererste: This is the case, when the url contains only 1 point. Since we don't have information about the index of this point, this assumes that this single point is the starting point.

src/stores/QueryStore.ts
Janekdererste: You are right. I changed the order so that ```SetPoint```, ```AddPoint```, ```SetParametersAtOnce``` are close to each other. I can put this back however.

src/NavBar.ts
easbar: Ah ok if I enter only one point in the url the first input field is populated and the second remains empty. So this seems to work :)

src/stores/QueryStore.ts
easbar: No no that makes sense then!

src/NavBar.ts
Janekdererste: Yes, that's what I meant :-)

PR #187, opened 2022-04-04
154 parse vehicle profile

### LINKED ISSUE #154: No routes are calculated for URLs missing the profile
Open this URL: https://graphhopper.com/maps2/?point=52.583872%2C13.311565&point=52.475677%2C13.437286

Note that the points are specified, but no profile is given. We see two markers, but no route is calculated. The default profile is chosen and even added to the URL. We then need to refresh the page (or similar) to calculate the route. Not sure if this is a bug or a feature? It certainly seems a bit inconsistent, because reloading the page does not give the same result.

Janekdererste:
fixes #154 . As discussed in the issue. The routing profile is now parsed from the url. This is done in two places. 
1. If the url has no profile property, the currently selected routing profile from the ```QueryStore``` is used
2. Since, on start up, the ```QueryStore``` might not have received a routing profile yet, the query store issues a routing request when it recevies a ```InfoReceived``` action. 

I also made sure stale routes are removed from the ```RouteStore``` when Points are removed. 

Most of the added lines are unit tests

### DIFF
--- a/src/NavBar.ts
+++ b/src/NavBar.ts
@@ -68,9 +68,12 @@ export default class NavBar {
             })
     }
 
-    private static parseProfile(url: URL) {
-        let profileKey = url.searchParams.get('profile')
-        return profileKey ? profileKey : ''
+    private static parseProfile(url: URL): string {
+        // we can cast to string since we test for presence before
+        if (url.searchParams.has('profile')) return url.searchParams.get('profile') as string
+        if (url.searchParams.has('vehicle')) return url.searchParams.get('vehicle') as string
+
+        return ''
     }
 
     private parseLayer(url: URL) {
--- a/src/NavBar.ts
+++ b/src/NavBar.ts
@@ -1,7 +1,7 @@
 import { coordinateToText } from '@/Converters'
 import { Bbox, RoutingProfile } from '@/api/graphhopper'
 import Dispatcher from '@/stores/Dispatcher'
-import { SelectMapStyle, SetInitialBBox, SetRoutingParametersAtOnce } from '@/actions/Actions'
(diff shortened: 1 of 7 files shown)


easbar:
Thanks! This works for me. The only thing I noticed is that when I enter e.g. `profile=xyz` in the url I briefly see an error message, that tells me there is no such profile, before the page is updated using the default profile (bike). If possible I think we should rather just show the error message? But then our app is probably in an inconsistent state, because there has to be some value for the profile?

Janekdererste:
> I briefly see an error message, that tells me there is no such profile, before the page is updated using the default profile (bike)

I think this is because the first routing request is issued with the profile from the url befor the info result comes in. We could delay the first routing request until the info result has returned. Also, on the ```InfoReceived``` action we could check wether the current profile is present in the available profiles from the info endpoint. If not we set the default value, otherwise we issue a route request with the one parsed from the url.

easbar:
I think the most important part is that the user notices that the requested profile is not present? Right now we only very briefly (just a short flickering really) see this error message and then the app shows a route for another vehicle (that we might not be interested in). So maybe we could just keep the error message and do not show a route at all? But what value should the profile menu show in this case? Maybe just an empty value? Users could then select another profile (but keep the coordinates) which would trigger another routing request. Is this what you meant also? I hope this does not complicate things too much, as it is only a small issue.

Janekdererste:
A little hacky, but the ```RoutingProfiles.tsx``` component could add the invalid selected profile to the ```<select>```. Then, if another option is selected, it disappears. 

grafik

easbar:
Nice! This is perfect :+1: I also like the little red flag icon :)

PR #188, opened 2022-04-12
Write and parse address in url

### LINKED ISSUE #10: Add addresses to url
Currently the url only stores the coordinates of selected points. It should also store the addresstext of points if present

Janekdererste:
fixes #10 

The addresses are written into the url inside the 'point' param. Separated from the coordinate with a '_' . I expect this character to appear in adresses not very often. The current approach handles a case where '_' would be part of an address by only parsing the part before the '_' character. 

Example URL: `http://localhost:3000/?point=52.464846%2C13.443016_Karl-Marx-Stra%C3%9Fe%2C+12057+Berlin%2C+Deutschland&profile=car&layer=OpenStreetMap`

### DIFF
--- a/src/NavBar.ts
+++ b/src/NavBar.ts
@@ -24,13 +24,18 @@ export default class NavBar {
         const result = new URL(baseUrl)
         queryStoreState.queryPoints
             .filter(point => point.isInitialized)
-            .map(point => coordinateToText(point.coordinate))
+            .map(point => this.pointToParam(point)) //coordinateToText(point.coordinate))
             .forEach(pointAsString => result.searchParams.append('point', pointAsString))
 
         result.searchParams.append('profile', queryStoreState.routingProfile.name)
         result.searchParams.append('layer', mapState.selectedStyle.name)
 
         return result
+    }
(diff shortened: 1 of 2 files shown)


Janekdererste:
I didn't tag you before, but maybe @karussell should also look at the url format, whether you are fine with this solution.

karussell:
> I expect this character to appear in adresses not very often. 

But even if it would be it wouldn't matter (?)

> The current approach handles a case where '_' would be part of an address by only parsing the part before the '_' character

I like it :+1: 

We might later introduce some kind of compression as for many points we will likely blow up the max URL length. (for current GH maps 1.0 we do not include the coordinates in the URL (only addresses), but this has other limitations like the need to wait for geocoding on the first route request and also in rare cases this produces different geocodes over time)

Janekdererste:
> But even if it would be it wouldn't matter (?)

The address wouldn't be properly added to the textbox. The coordinate would still be correct. I stopped there in favor of a 'good enough solution'

Mergin this, becuase of your thumbs up.

PR #189, opened 2022-04-13
Icons for routing profile

### LINKED ISSUE #58: Use icons instead of emojis
I see a strange gender (?) symbol next to the vehicle icons:

image
image
image

but this only happens in Firefox. Using Chrome it looks normal, for example:

image

Janekdererste:
This fixes #58 

This puts the icons for routing profiles on  top of the search panel. 

@karussell The old graphhopper maps repository didn't contain svg icons for all the routing modes. To get a first impression I used fontawesome Icons for the missing ones. This looks a little weird. What would you like to do?

grafik

TODO:
- [ ] Unify Icons


karussell:
Nice - thanks!

> What would you like to do?

I will search if I can find the svg versions somewhere. So, bike and car is missing?

Janekdererste:
> I will search if I can find the svg versions somewhere. So, bike and car is missing?

Yes, plus wheelchair, which had syntax errors, which would also cause hickups in the browser, 
Also, some vehicles are not all transparent. The wheels of the truck and the windows of the bus for example

bt90:
@Janekdererste could you add `width="24" height="24"` to bike.svg and foot.svg for the sake of consistency?

Issue #150, opened 2021-11-03
include points additionally to route in bounding box


karussell:
is a bug and was part of #53

PR #192, opened 2022-06-03
Grow BBox due to query points

karussell:
For #150 and #53.

When a too small bbox was found (e.g. slightly below 0.001) it can grow to nearly 0.002. Should we set it always to 0.001? (makes the code a bit less understandable and found the current behaviour also fine)


Janekdererste:
This looks much better than in the current master branch.

Issue #163, opened 2021-11-18
Do not zoom to route bbox when moving markers


easbar:
When we zoom in and move a marker we don't want to automatically zoom out to the route view again. The camera should just remain fixed like in GH maps 1.

easbar:
apparently this was fixed, somewhere...

Issue #177, opened 2022-01-20
Do not zoom to route bounding box unconditionally


easbar:
Currently we change the camera position such that a route fits into the window whenever a route request succeeds. However, right now there are two cases coming to my mind where this is not ideal:

1) We already did an estimation of the bounding box from the given url points and updating the camera position creates a little flickering effect even though the route would be entirely visible anyway

2) We already updated the map bounds for the first /route request, but when the second /route request (that includes alternative routes) succeeds we update the map bounds again. This is especially annoying when the alternative routes take some time and the user already started zooming/panning the map. E.g. load this URL, pan the map and you will see it jumps back at some point: http://localhost:3000/?point=48.988641%2C7.180233&point=50.856244%2C12.297411&profile=foot&layer=OpenStreetMap

Nothing major, but keeping this as a reminder here.

PR #193, opened 2022-06-05
do not zoom if marker was dragged

karussell:
The current behaviour is pretty annoying and so I disabled this, in this PR.

(not sure if this is the proper way to do this or if we should have the zoom parameter e.g. in SetPoint)

Related to #177

### DIFF
--- a/src/layers/UseQueryPointsLayer.tsx
+++ b/src/layers/UseQueryPointsLayer.tsx
@@ -87,11 +87,14 @@ function addDragInteractions(map: Map, queryPointsLayer: VectorLayer<any>) {
         const coordinateLonLat = toLonLat(feature.getGeometry().getCoordinates())
         const coordinate = { lng: coordinateLonLat[0], lat: coordinateLonLat[1] }
         Dispatcher.dispatch(
-            new SetPoint({
-                ...point,
-                coordinate,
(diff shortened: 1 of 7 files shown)


easbar:
Cool, thanks! This fixed #163, right? Makes drawing markers a lot more enjoyable for sure :)

Putting the `zoom` parameter into `SetPoint` makes sense to me. The only thing I was wondering is if the `QueryStoreState` and the `RoutingArgs` are the right place for the `zoom` parameter, because it is only related to our UI code and not the routing request. I mean I can see how it works this way and maybe it's not too important, but it seems a bit odd, so I was wondering if there was another way to do this?

Janekdererste:
> The only thing I was wondering is if the QueryStoreState and the RoutingArgs are the right place for the zoom parameter

I had the same thought, but then I didn't have another idea quickly. Since we all don't feel good about this we could think of a better way to do this, or at least remember it for the next UI related property and do it then.

easbar:
Ok, good idea. I added a todo comment here: e95f88eb45b2846fed3e24dfd734e19fce2bd666

The zoom property is like a parameter that describes the UI parameter, but then it is also only 'valid' for a specific request: When dragging the marker it is set to false **for the next request**, but when we type two cities in the query boxes directly it should be true for the next request...

PR #198, opened 2022-06-25
minor profile style fix

karussell:
The "outline" circle of the button is not fully visible on firefox and chromium. And on chromium it is black.
 
Before Firefox:

image

After Firefox:

image

Before Chromium:

image

After Chromium:

image

Unfortunately for Chromium the first profile is now even less aligned with the input. Not sure if this is worse and how to fix this. We could probably reduce `padding: 0.45rem;` again to reduce this problem but still it is no 100% fix.


easbar:
Strange, on my laptop the circle is fully visible in FF and not black in Chrome either. Or actually... there is no outer circle at all?

image

image

easbar:
Ok, I see. Same problem here then.

Issue #206, opened 2022-09-02
custom model box improvements


karussell:
Important:

 - [x] we need an "apply" button or similar as not every key stroke can issue a request (or not everytime the model is valid) -> i.e. an explicit apply button would make it more transparent (we can also disable the button if a request is going on)
 - [x] the route calculation with a custom_model is complex and so we should disable alternative calculation for requests where the points have a longer distance than e.g. 100km. The entire request should be rejected if larger than e.g. 700km or similar
 - [x] read custom_model from URL parameter

Optional:

 - [ ] somehow allow comments in the custom model? with them we could explain more complex models but also indicate a simple help for the default example e.g. that Ctrl+Enter or Ctrl+Space work etc. Or that the custom_mode is applied on top of the selected routing profile and what the entries "speed" and "priority" roughly do.

PR #208, opened 2022-09-03
Allow setting an initial custom model via an url parameter

easbar:
For #206 

With this PR we can append an arbitrary string (but better a valid custom model json string) via `&custom_model={"whatever": "here"}` to the url. This will start the app with an open custom model box containing this string.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -124,7 +124,11 @@ function LargeScreenLayout({ query, route, map, error, mapOptions, info }: Layou
                         selectedProfile={query.routingProfile}
                         autofocus={false}
                     />
-                    <CustomModelBox enabled={query.customModelEnabled} encodedValues={info.encoded_values} />
+                    <CustomModelBox
+                        enabled={query.customModelEnabled}
+                        encodedValues={info.encoded_values}
+                        initialCustomModelStr={query.initialCustomModelStr}
+                    />
                     <div>{!error.isDismissed && <ErrorMessage error={error} />}</div>
                     <div className={styles.routingResult}>
                         <RoutingResults
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -25,12 +25,16 @@ import { getApi, setApi } from '@/api/Api'
 import MapActionReceiver from '@/stores/MapActionReceiver'
 import { createMap, getMap, setMap } from '@/map/map'
 
-let locale = new URL(window.location.href).searchParams.get('locale')
+const url = new URL(window.location.href)
+const locale = url.searchParams.get('locale')
 setTranslation(locale || navigator.language)
 
-// set up state management
-setApi(config.api, getApiKey())
-const queryStore = new QueryStore(getApi())
+// use graphhopper api key from url or try using one from the config
+const apiKey = url.searchParams.has('key') ? url.searchParams.get('key') : config.keys.graphhopper
+setApi(config.api, apiKey)
+
+const initialCustomModelStr = url.searchParams.get('custom_model')
+const queryStore = new QueryStore(getApi(), initialCustomModelStr)
 const routeStore = new RouteStore(queryStore)
 
 setStores({
@@ -71,9 +75,3 @@ root.style.height = '100%'
 document.body.appendChild(root)
 
 ReactDOM.render(<App />, root)
-
-function getApiKey() {
-    const url = new URL(window.location.href)
-    // use graphhopper api key from url or try using one from the config
-    return url.searchParams.has('key') ? url.searchParams.get('key') : config.keys.graphhopper
-}
--- a/src/sidebar/CustomModelBox.tsx
+++ b/src/sidebar/CustomModelBox.tsx
@@ -78,6 +78,7 @@ export default function CustomModelBox({ enabled, encodedValues, initialCustomMo
         const instance = create({}, (element: Node) => divElement.current?.appendChild(element))
         setEditor(instance)
 
+        instance.cm.setSize('100%', '100%')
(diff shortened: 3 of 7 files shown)


src/App.tsx
```diff
@@ -124,7 +124,11 @@ function LargeScreenLayout({ query, route, map, error, mapOptions, info }: Layou
                         selectedProfile={query.routingProfile}
                         autofocus={false}
                     />
-                    <CustomModelBox enabled={query.customModelEnabled} encodedValues={info.encoded_values} />
+                    <CustomModelBox
+                        enabled={query.customModelEnabled}
+                        encodedValues={info.encoded_values}
+                        initialCustomModelStr={query.initialCustomModelStr}
```
easbar, line 130: Note that the `initialCustomModelStr` prop is really what it says: The **initial** string value for the custom model box. It is only ever used when the component is created initially, see also my comments here: https://github.com/graphhopper/graphhopper-maps/pull/183#discussion_r806256275. Not sure if there is a better way to make this clear.

src/index.tsx
```diff
@@ -25,12 +25,16 @@ import { getApi, setApi } from '@/api/Api'
...
 
-// set up state management
-setApi(config.api, getApiKey())
-const queryStore = new QueryStore(getApi())
+// use graphhopper api key from url or try using one from the config
+const apiKey = url.searchParams.has('key') ? url.searchParams.get('key') : config.keys.graphhopper
+setApi(config.api, apiKey)
+
+const initialCustomModelStr = url.searchParams.get('custom_model')
+const queryStore = new QueryStore(getApi(), initialCustomModelStr)
```
easbar, line 37: With #207 we can now pass the initial value directly to the store instead of creating the store and then updating it's state immediately.

src/sidebar/CustomModelBox.tsx
```diff
@@ -76,9 +77,16 @@ export default function CustomModelBox({ enabled, encodedValues }: CustomModelBo
...
         setEditor(instance)
-        setCustomModelExample(instance, examples['empty'])
-        // todo: minor glitch: if the initial model is invalid we see an 'Invalid custom model' error notification, even
-        //       though the custom model box is closed initially
+
+        if (initialCustomModelStr != null) {
+            // prettify the custom model if it can be parsed or leave it as is otherwise
+            try {
+                initialCustomModelStr = customModel2prettyString(JSON.parse(initialCustomModelStr))
+            } catch (e) {}
```
easbar, line 85: In our initial implementation (some commit in #183) the app crashed when the string was no valid json. Now we catch this error here and simply paste it as given to the custom model box.

src/stores/QueryStore.ts
```diff
@@ -95,10 +97,11 @@ export default class QueryStore extends Store<QueryStoreState> {
             routingProfile: {
                 name: '',
             },
-            customModelEnabled: false,
+            customModelEnabled: initialCustomModelStr != null,
```
easbar, line 100: We open the custom model box if the `custom_model` url parameter is given. There is still something wrong with the css, because the box overlaps the 'Help' button and the 'Examples' Dropdown in this case. @karussell do you have an idea how to fix this?

src/index.tsx
```diff
@@ -25,12 +25,16 @@ import { getApi, setApi } from '@/api/Api'
...
 setTranslation(locale || navigator.language)
 
-// set up state management
-setApi(config.api, getApiKey())
-const queryStore = new QueryStore(getApi())
+// use graphhopper api key from url or try using one from the config
+const apiKey = url.searchParams.has('key') ? url.searchParams.get('key') : config.keys.graphhopper
+setApi(config.api, apiKey)
+
+const initialCustomModelStr = url.searchParams.get('custom_model')
```
easbar, line 36: We take the custom model from the url, but then we set the url from our state somewhere, so the custom_model parameter disappears. Should we add it to the url parameters? And even keep it updated the whole time? Not sure...

src/stores/QueryStore.ts
easbar: I use this url: http://localhost:3000/?custom_model={"distance_influence":70}
Then it looks like this (in Chrome, and the same problem in Firefox): 

image

When I close and open again the custom model box it no longer overlaps the buttons.

src/index.tsx
easbar: Looks like this is easy (?): c96eda0d9ec89f8bc72f9e352f89bcc155d3b5af
Can you check if this is what you want? And maybe we need some url encoding or something?

src/stores/QueryStore.ts
easbar: > But does not happen for me in Firefox :D

oh no :( not this again. For me it's the same in Chrome and FF

src/App.tsx
easbar: > Couldn't we avoid this parameter somehow? 

To put it short: I don't really think so.

> similar to zoom it does not seem to fit well into QueryStore

Yes, but just because we did not create a separate store for this (and maybe zoom) that might be a better fit it does not say we don't need it.

> E.g. we could parse the custom_model URL parameter before we set the QueryStore.custom_model variable and if this parsing fails set it to null? 

The custom_model variable in QueryStore is not the backend for the editor content (see the comment I linked above).

src/stores/QueryStore.ts
easbar: Yes, that fixed it!

easbar: ... well not completely, because the box remains blank when there is no url parameter after opening the box, but see latest commits.

PR #210, opened 2022-09-04
Add routing graph layer

### LINKED ISSUE #24: routing graph as vector tiles (only for development)

easbar:
Fixes #24.

Should be quite similar to what we had in GH maps 1. Maybe the biggest difference is that I put the road details into popup shown on top of the map. To enable the feature we need to set `routingGraphLayerAllowed: true` in `config-local.js`.

The popup looks like this currently: 

image

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -8,6 +8,7 @@ import {
     getErrorStore,
     getMapOptionsStore,
     getPathDetailsStore,
+    getMapFeatureStore,
     getQueryStore,
     getRouteStore,
     setStores,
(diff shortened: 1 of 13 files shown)


src/index.tsx
```diff
@@ -24,6 +25,7 @@ import * as config from 'config'
 import { getApi, setApi } from '@/api/Api'
 import MapActionReceiver from '@/stores/MapActionReceiver'
 import { createMap, getMap, setMap } from '@/map/map'
+import MapFeatureStore from '@/stores/MapFeatureStore'
```
easbar, line 28: I did not know where else to put the coordinate and contents of the hover popup, so I added a new store that we can maybe use for other similar things as well...

easbar:
Currently the graph layer is (partly) hidden when a vector layer is used as background layer, but maybe we can ignore this for now as our focus lies on raster backgrounds.

Btw: In a previous version I tried showing the graph layer above the vector background, but below the map labels. I was able to identify the label layers by checking `layer.type === 'symbol'`.

easbar:
> It reads as you did not do this but in my case the labels are already above the routing graph :) (!?)

Yes, apparently they are. But why? Oh, well :)

> minor thing: it would be nice to enable the layer also when clicking on the label but I couldn't fix it quickly as the e.target.checked is not available in the parent.

Good idea. I fixed this, see last commit.

PR #211, opened 2022-09-05
avoid overloading of server when using custom_model

karussell:
For #206.

Please have a look @easbar as I'm unsure if this really belongs to QueryStore. I tried to move it to CustomMobelBox but it wasn't that nice too (i.e. I tried to avoid making the custom model valid in the first place if distance is too big).

### DIFF
--- a/src/stores/QueryStore.ts
+++ b/src/stores/QueryStore.ts
@@ -253,34 +253,40 @@ export default class QueryStore extends Store<QueryStoreState> {
 
     private routeIfReady(state: QueryStoreState): QueryStoreState {
         if (QueryStore.isReadyToRoute(state)) {
-            const requests = [
-                QueryStore.buildRouteRequest({
-                    ...state,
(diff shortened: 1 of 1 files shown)


src/stores/QueryStore.ts
```diff
@@ -258,7 +260,26 @@ export default class QueryStore extends Store<QueryStoreState> {
                 }),
             ]
 
-            if (state.queryPoints.length === 2 && state.maxAlternativeRoutes > 1) {
+            let allowAlternatives = state.queryPoints.length === 2 && state.maxAlternativeRoutes > 1
+            if (state.customModelEnabled && state.queryPoints) {
```
easbar, line 264: What is the `state.queryPoints` check for?

easbar: To me it looks like this is already covered in `isReadyToRoute`

easbar:
LGTM thanks!

PR #213, opened 2022-09-05
apply button for custom model 

karussell:
For #206.

This makes the handling with a custom model more obvious.

The button is disabled while the query is ongoing and also if the model is invalid (and so I removed the error message or shall we keep it?).

Additionally the settings button is moved to the right side. This is not only more compact but reduces misclicks.


src/sidebar/CustomModelBox.tsx
```diff
@@ -89,7 +92,6 @@ export default function CustomModelBox({ enabled, encodedValues, initialCustomMo
             initialCustomModelStr == null ? customModel2prettyString(examples['empty']) : initialCustomModelStr
 
         instance.validListener = (valid: boolean) => {
-            dispatchCustomModel(instance.value, valid)
```
easbar, line 92: Why did you remove this?

src/sidebar/CustomModelBox.tsx
```diff
@@ -129,6 +131,11 @@ export default function CustomModelBox({ enabled, encodedValues, initialCustomMo
         [editor, isValid]
     )
 
+    const isQueryOngoing = () => {
+        if (!getQueryStore().state.currentRequest.subRequests[0]) return false
+        return getQueryStore().state.currentRequest.subRequests[0].state == RequestState.SENT
```
easbar, line 136: Is this the same as `getQueryStore().state.currentRequest.subRequests[0]?.state === RequestState.SENT`?

easbar: ... and not sure if it is a good idea to access the state like this, maybe we can pass this boolean into CustomModelBox as a prop? I don't know, just wondering.

src/sidebar/CustomModelBox.tsx
easbar: > Why?

Because you removed the code that updated the state :)

> I wanted to send the request only if the Apply button was clicked (like in GH 1) not when the custom_model was changed.

Yeah, but calling the dispatch method here is not equal to sending a request.

easbar: > I thought, that this always triggered a request whenever the model was valid, which was a bad experience.

No, that would be a bad experience, so in master no request is triggered and a request is only triggered when we press enter in the input box etc.

> When is a request triggered?

When `QueryStore.isReadyToRoute` is true.

> Ok, probably this is something I do not yet fully understand :)

This might be hard to miss, but the `dispatchCustomModel` method has a by default false boolean parameter `withRouteRequest`, but even when this is true a request is only triggered when `isReadyToRoute`...

> And from experiments this worked as expected.

Try moving markers and entering and deleting stuff in the input boxes and it won't work as expected.

easbar: There are just so many cases:

* a custom model might be not parseable, so we cannot even go from a string to javascript object
* a custom model might be invalid, so we want to grey out the apply button (that previously did not even exist)
* points might be missing so we don't want to request either
* the custom model box could be disabled, but enabling it again we want to see the model we had before

etc.

easbar: The custom model box' task is only to maintain these two variables: `QueryStoreState.customModel` and `QueryStoreState.customModelValid`. All the other interactions like clicking the apply button should not mess with this at all and only trigger the `routeIfReady` code which in turn just reads these variables. Then it won't be very complicated I hope.

easbar: > Ok, I see. And including back dispatchCustomModel does not fix the problem?

I'm not sure yet, but I try to understand this whole thing again and maybe it can be simplified a bit.

easbar: I had the same thought, but I'm not sure yet.

Probably still not perfect, but hopefully this starts to make it more clear: https://github.com/graphhopper/graphhopper-maps/pull/213/commits/779c03a343a22cfe6fe507bb250bd97e2983a0bd

easbar: One remaining problem I found is this:

* http://localhost:3000/?point=51.646173%2C10.458679&point=51.613779%2C10.491638&profile=bike&layer=Omniscale
* enter some garbage
* move a marker -> the route disappears (makes sense)
* close the box -> we get a route (without custom model, makes sense)
* open the box again -> normally the route is re-calculated, but in this case since the custom model is garbage (null or not valid) `routeIfReady` does nothing, so we keep seeing the **old** route (without custom model). It might be more consistent to delete the route in this case, similar to the `InvalidatePoint` action?

easbar: 579e626be02fe2cf31b25143bcf63367d7b5fb75

I like this better (?)

easbar: I now also send a request from the url parameter when the app is loaded for the first time: 9eede15ec9baa6a0a81b09d243bf9aa0240a2347

src/sidebar/CustomModelBox.tsx
easbar: I think it would be the first time we do it differently, so better not. I changed this here and it seems to work: 686e97919fe03882e0423ee7650bf8539fe7537c

PR #214, opened 2022-09-05
Custom model limit, clean up (?) if/else logic

easbar:
I don't know if this is just me, but I found the if/else logic hard to read. Does this code do what you intended and do you consider it is it an improvement? As far as I understood we want to:

* reject queries with custom models > 500km
* disable alternatives with custom models > 200km


easbar:
Ok, but what is the `200_000` meter limit for then? I thought you want to use alternatives below 200_000, not use alternatives between 200_000 and 500_000 and do not send a request at all above 500_000.

easbar:
like this?

easbar:
Sorry yes sure my bad.

easbar:
> Probably I should have written some simple tests to make refactoring easier :)

better just straightforward code that does not require thinking :)

Issue #203, opened 2022-08-31
Context menu breaks when switching between small and large screen layout


easbar:
1) right click the map -> context menu opens
2) close the context menu (using the new x button, or choose something like 'Center map here')
3) press F12 to open the dev tools
4) right click the map again -> the default browser context menu opens (Save page as... etc.), we can no longer open the GH maps context menu

easbar:
Apparently this was fixed somewhere, or at least I can no longer reproduce this.

easbar:
This still happens when I 

1) open the app
2) reduce the window size until it switches to the small screen layout

And it even happens in Chrome too.

easbar:
I think this is because we register the click listeners to the container element, which changes once we switch between the small & large layout:

https://github.com/graphhopper/graphhopper-maps/blob/de11ef9df3f00ee7baea5e86bbcc38fbf269c498/src/layers/ContextMenu.tsx#L40-L58

easbar:
> Changing map.once to map.on seems to fix it. I'm just not sure if we need to clean up the listeners somehow then?

Apparently no. I used this function that lists all event listeners and switched there and back between the small and the large layout. The number of listeners did not increase with every switch: https://gist.github.com/tkafka/1c5174ed5b446de7dfa4c04a3b09c95f

PR #232, opened 2022-10-04
context menu: below cursor + without close button + with icons

karussell:
The problem was when you are zoomed out then the popup is too big if you clicked into an area north of Germany. And you first have to zoom in to click on "start". So moving the popup right-below the cursor seems to feel better and additionally we removed some code :)

(the elevation popup above the route is left unchanged)

Related to #136 and #135.

### DIFF
--- a/src/layers/ContextMenu.module.css
+++ b/src/layers/ContextMenu.module.css
@@ -1,6 +1,6 @@
 .popup {
     background-color: white;
     box-shadow: 0 1px 4px rgba(0, 0, 0, 0.2);
-    padding: 12px 0 8px 0;
+    padding: 0.5em 0;
     border-radius: 3px 3px 3px 3px;
 }
--- a/src/map/Popup.module.css
+++ b/src/map/Popup.module.css
@@ -19,31 +19,7 @@ button.entry div {
     cursor: pointer;
     outline: inherit;
     font-size: 16px;
-    padding: 3px 22px 5px 17px;
-}
-
-.closeBtn:hover {
-    cursor: pointer;
-    color: black;
-}
-.closeBtn {
-    color: gray;
-}
-.closeBtn,
-.closeBtn:hover {
-    border: none;
-    background: none;
-    position: absolute;
-    top: 0.2em;
-    right: 0.2em;
-
-    width: 20px;
-    align-self: flex-start;
-    flex-shrink: 0;
-}
-
-.closeBtn svg {
-    width: 100%;
+    padding: 0.15em 0.8em;
 }
 
 .entry:hover {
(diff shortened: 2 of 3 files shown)


easbar:
I liked the little 'arrow' pointing at the clicked location:

image

vs. now

image

can we put this back (I have no idea where it came from)?

easbar:
Ok in maps 1 we did not have this 'arrow' either... Still in maps one the menu looks better I think:

image

image

easbar:
> better?

I like the green/blue/red icons for from/via/to. The old menu still looks better to me, maybe because of the smaller line spacing, less padding and smaller corner radius?  But probably not so important. And of course we should get rid of the ugly 'x' button at some point.

> they are unnecessary because all the other corners are rounded and the arrowhead is formed indirectly :)

yeah.. but that looks a bit strange :D

> Interesting: if you are at the bottom of the map the map moves so that the popup is then completely visible :)

That is probably an OL feature? And maybe it does not work properly in master, because of the css you removed here?

> And can you check if you still see https://github.com/graphhopper/graphhopper-maps/issues/203?

I cannot reproduce this here, but I cannot reproduce it in master, either.

easbar:
> So you want to look the context menu exactly like the old I guess :) 

I don't know, I was just saying the old one looks better to me. And actually I don't even think it fits very well with the rest of the app:

image

* The sidebar has a horizontal grey line above 'Powered by GraphHopper', but the menu has a *different* horizontal grey line (its thicker)

* The padding on the sides is bigger in the context menu than in the sidebar, even though the sidebar is wider overall, so the relative padding is much bigger in the menu.

* The vertical spacing is hard to compare, so if we made it smaller in the menu it would still 'match' the look of the other components?

* The 'x' button in the menu looks different than the 'x' button of the elevation diagram

* The corner radius of the elevation diagram looks smaller than the one of the menu, and maybe even the one we use for the sidebar is smaller thane the one of the menu?

image

* The font sizes seem to be different in the elevation diagram ('Elevation' and 'Legend') and in the instructions list.

easbar:
... but maybe this is for another issue :) https://github.com/graphhopper/graphhopper-maps/issues/136

easbar:
> Hm, you mean a smaller component should get less padding?

No, I thought it should probably be the same in absolute terms, but a smaller component should definitely not have a bigger (absolute) padding?

> what do you mean here?

I was trying to say: Yes, we should try to match the look of the context menu with the other components, but since I could not see anything that was similar to the menu I thought we can freely choose how much empty room there is between the map entries (rows).

> The elevation diagram had the same 10px border radius :) 

Really? In my above screenshot it looks like the elevation diagram radius is bigger?!

easbar:
> Can I merge or further improvements possible :) ?

You can always merge and always further improvements are possible :D

easbar:
Thanks for this!

Issue #216, opened 2022-09-05
Our app does not work in React StrictMode


easbar:
see: https://github.com/graphhopper/graphhopper-maps/commit/40238bfde7fb60d19d73ddd9bd471f83c555636b

For example when StrictMode is enabled the map's context menu no longer works and the positioning of the custom model editor is broken.

easbar:
I think the reason the context menu does not open in strict mode is that the context menu listener is removed here: 

https://github.com/graphhopper/graphhopper-maps/blob/d37667728b273bf026fd9305acb0fa9a8c450327/src/layers/ContextMenu.tsx#L55

but it is only added via the `map.once('target:change')` callback here:

https://github.com/graphhopper/graphhopper-maps/blob/d37667728b273bf026fd9305acb0fa9a8c450327/src/layers/ContextMenu.tsx#L39

Since StrictMode runs the useEffect multiple times the listener is removed but not added again. And the reason we add the listener in this strange way, is that the useEffect is called for the `map` before it has a target element. Maybe we should do this another way, but currently I don't really know how. The target element is attached here at the moment:

https://github.com/graphhopper/graphhopper-maps/blob/d37667728b273bf026fd9305acb0fa9a8c450327/src/map/MapComponent.tsx#L14

PR #235, opened 2022-10-05
Close context menu when we click the map

### LINKED ISSUE #135: A left click in the map should close the context menu

easbar:
Fixes #135.

Apparently all we need to do is add a click handler to the map, but for some reason we cannot call `setMenuCoordinate(null)` as @karussell figured out in #232.

When I tried to fix the issue here it never worked as expected:

* After I added the click handler via `map.getTargetElement().addEventListener('click', ...` the menu closed when clicking the map, but it also just closed when I clicked on one of the menu entries (without any further action)
* I was then able to fix this with this workaround in Popup.tsx in 43d28bc0532831d3d1546f572b193f19204d9621
* ... but on iOS the menu would always close when I let go the long press, unless I was long-pressing and dragging
* When I tried `map.on('click'...)` instead it worked in the browser even without the change in Popup.tsx, but I had the same dragging issue on iOS again.
* What does seem to fix it is not calling `setMenuCoordinate(null)`, but why?

### DIFF
--- a/src/layers/ContextMenu.tsx
+++ b/src/layers/ContextMenu.tsx
@@ -45,7 +45,7 @@ export default function ContextMenu({ map, route, queryPoints }: ContextMenuProp
             map.getTargetElement().addEventListener('touchmove', () => longTouchHandler.onTouchEnd())
             map.getTargetElement().addEventListener('touchend', () => longTouchHandler.onTouchEnd())
 
-            map.getTargetElement().addEventListener('click', () => {
+            map.on('click', () => {
                 overlay?.setPosition(undefined)
                 // setMenuCoordinate(null)
             })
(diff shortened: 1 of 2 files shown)


easbar:
I checked again:

```
            map.on('click', () => {
                overlay?.setPosition(undefined)
            })
```

works on my laptop, but does not work on iOS: When I do the long press the menu opens, but then closes again as soon as I let go, unless I do an additional drag.

```
            map.getTargetElement().addEventListener('click', () => {
                overlay?.setPosition(undefined)
                setMenuCoordinate(null)
            })
```

Does not work on my laptop as clicking on one of the menu entries just closes the menu, at least without the workaround in Popup.tsx

```
            map.getTargetElement().addEventListener('click', () => {
                overlay?.setPosition(undefined)
            })
```

Seems to work, so we have to set the event listener on the target element **and** not call `setMenuCoordinate(null)`.

easbar:
Can you also try this? http://gh-maps-react.s3-website.eu-central-1.amazonaws.com/cleanup_context_menu
Apparently this also works in strict mode (compared to master and the solution here)

easbar:
Anyway, I'll merge this so you can also merge your context menu changes and the above version is maybe something for #216. But just the fact that here something went wrong when we called `setMenuCoordinate(null)` and that it does not work in strict mode tells me we are still doing something wrong :)

easbar:
>  or will it close in these cases too?

Probably not, but let's just try again when it is merged.

PR #240, opened 2022-10-10
Move gear button from CustomModelBox into Search

karussell:
The "gear" button to open the custom model box belongs technically to the CustomModelBox. But this makes problems with the CSS and we hacked it together so that it is horizontally aligned with the "add location" button and vertically aligned with the right border. Unfortunately this does not properly work if we add more locations as the "minus" buttons that appear are currently not vertically aligned with the gear button. And this will continue to make problems if we e.g. introduce new buttons on the same line (of the "add location" button) like an info/help/imprint and share URL button or similar. 

This is the reason I moved the gear button to the Search component which makes the layout handling clean although now the Search component depends on some custom model box state (customModelEnabled).


easbar:
I wonder if it not rather belongs into the row with the profile icons. Not just for layout reasons, but also because it is about changing the profile.

easbar:
Thanks :)

PR #244, opened 2022-10-11
Push the bundled app to npmjs.org

easbar:
This PR adds a GitHub workflow that builds the bundled application in the 'dist' folder and releases it on npmjs.org.


.github/workflows/push_dist_to_npm.yml
```diff
@@ -0,0 +1,31 @@
+name: Push dist to npm
+on: push
```
easbar, line 2: For now this runs on all branches, but once the integration with the GraphHopper server works we should probably limit this to the main branch

.github/workflows/push_dist_to_npm.yml
```diff
@@ -0,0 +1,31 @@
...
+      - name: Setup Node.js
+        uses: actions/setup-node@v3
+        with:
+          registry-url: https://registry.npmjs.org/
+          node-version: lts/gallium
+
+      - name: Build
+
+        run: |
+          npm version --git-tag-version=false 0.0.0-$(git rev-parse HEAD)
```
easbar, line 22: npm requires a certain version string format. But since we want to use the git sha as version I set this to `0.0.0-<git sha>`.

.github/workflows/push_dist_to_npm.yml
```diff
@@ -0,0 +1,31 @@
...
+          # limit distribution to the dist folder (just for this build)
+          node -e "const packageJson=require('./package.json'); packageJson.files=['dist/']; require('fs').writeFileSync('package.json', JSON.stringify(packageJson, null, 4));"
```
easbar, line 24: This build only releases the bundled/minified production app, not all the source code.

PR #251, opened 2022-10-12
Allow switching to miles/feet

### LINKED ISSUE #21: miles vs. kilometer config

easbar:
Fixes #21. Compared to maps1 it also changes the units in the path detail/elevation diagram.

todo: 

* [ ] For now I simply added an ugly 'toggle units' button next to the 'powered by' logo. Where should we put this button? Or shall we derive this from the locale? Or (only) make it configurable via config.js?


easbar:
I think what we did before worked pretty well?

image

Now that we show multiple alternative routes, maybe we should only show this for the selected route? There is some space left above the detail&export buttons:

image

?

easbar:
> btw: is the zoom parameter also a candidate for this settings store?

Yes, maybe. What does it do? But if yes, let's move it afterwards.

> We could make this automatic (like US=miles and rest is km) 

Yes, choosing the default based on the locale seems useful, but maybe this is even something for a follow-up.

> and make the distance unit clickable? (Hm, probably not as this is too tricky on mobile?)

Also the distance appears in multiple places and it would be ugly to show something like `km|mi` everywhere.

> Yes, we could introduce an options menu but it is very similar to the "customization" gear button. Not sure if we should do this.

Yes, the gear button totally looks like the place where to change settings like this and the custom model editor is rather unexpected when we click it... Maybe it should just be a text label 'Custom Model'. For example Google Maps also uses an 'Options' text label (no icon). But that is probably for another issue ("it is hard to understand that the gear icon opens the custom model editor").

> Or change the meaning of the gear button (to just open the options panel) but then have different logic for custom model enabling e.g. a toggle button inside this options panel. And in the options panel the custom model box could be still visible but with a style that indicates when it is enable or disabled. Shall I prototype something like this?

Sure, feel free to try. But not sure if we should combine the settings and the custom model editor, because changing the distance units in a menu would be like opening the menu, changing the setting and closing it again, while the custom model editor is something that rather should stay open when users decide to use it.

> Not only a bit ugly but also again it could be tricky on mobile?

To me this seems to be the best (and easiest) option so far. Clicking the `km | mi` button would not be more or less tricky than clicking details or export? Or maybe even better: Google Maps also has this menu button: 

image

And if we had a menu like this we could also put stuff like the link to the attribution page etc. there.

easbar:
Ok, maybe just go ahead, I don't have a complete solution on my mind currently.

easbar:
> And we would have to differentiate the "applied" custom model box from the "visible" custom model box (or close the box when it is applied, like a "save" button and have another pinned button to explicitly keep it open). But this is probably for another issue :)

Thanks for trying! And yes in this case we would probably have to keep things separated a bit: One button that only determines whether or not the routing settings panel is open or not. And within the settings panel there would have to be some checkbox or save/apply button that determines whether the custom model is active or not. Something for another issue.

> What about your idea of having the customization on the profiles line but move this profiles below the locations? Then there is space for km/mi:

Yes, sure that works for me. I think I would slightly prefer the gear button on the right and underlining the selected profile then. Or maybe just remove the 'add' icon?

> btw: I also thought about the hamburger menu but all the apps I tried slide in a new panel over the existing stuff and this would be ugly in general IMO and especially when the custom model box is kept open

Yes sliding on top seems ugly, but if they all do it it might be ok? Google Maps also greys out the rest of the screen when the menu is open. An alternative could be some kind of 'tabbing' where the look of the sidebar stays the same, but its contents are changed entirely. Something like this could be useful anyway if we wanted to support isochrones, map matching or advanced rendering options for the mvt graph.

easbar:
Shall I try to rearrange the components as you suggested or did you do this already or want to do it yourself?

easbar:
> Yes, and if we close the settings button and the custom model is active we would still need to indicate this via a green dot or something and then also show the calculation progress bar around the settings button.

Ah yes the progress bar. Could we show the progress for every request, not just when custom models are enabled (assuming it goes so fast it won't be visible usually)? Or maybe only show it after a certain delay)? And yes some indication it is enabled would also be required.

> I tried it already and it feels also a bit ugly :D ... the problem with the profiles below the locations is that the more locations you have the more you move the profiles out of the visible range.

Ok, I see.

> Btw: moving the custom model box above the locations so that we can keep the profiles panel above the locations is not that bad too from my simple trials. Especially when we can close the custom model box and still keep it enabled.

Yes and maybe this way it is easier to understand that the model box is about tuning the profile.

> I think I will just put a km button left to the gear button (i.e. only show the active unit without | mi but with a tooltip so that it is as small as a button) and then we are done for now. 

Ok sure. 

> Probably we should only show this for lang=en (?)

Ok yes

easbar:
Unrelated to this issue...

> And yes some indication it is enabled would also be required.

Maybe the profile selection should also be separated from the 'base profile' selection of the custom model (at least at some point). In this case the current profile selection would be about selecting either one of the prepared/built-in profiles like car, bike and so on or a custom profile. And that the custom profile is selected will be indicated by the blue underline of the 'custom profile icon' (whatever that might be, like the gear icon we use currently). And when the custom profile box is open there would be some way to choose the base profile for the custom model, and maybe this even becomes part of the custom model JSON? But for this we first need to figure out what 'base profiles' really are...

PR #255, opened 2022-10-15
Debug profile list bug

### LINKED ISSUE #254: Profile list is empty on iOS
Currently we fetch the /info endpoint with the profiles right at the beginning here: 

https://github.com/graphhopper/graphhopper-maps/blob/9a911c2e6caaba16f6bfb8466dbb934afbcfe43b/src/index.tsx#L69

We already registered the stores at the Dispatcher at this point so this looks ok.

However, we do not subscribe to the ApiInfoStore until the render function in `App.tsx` is called for the first time here:

https://github.com/graphhopper/graphhopper-maps/blob/9a911c2e6caaba16f6bfb8466dbb934afbcfe43b/src/App.tsx#L64

So apparently it can happen that the /info request returns and the profiles are updated in ApiInfoStore before we registered the `setInfo` state update function at the ApiInfoStore. In this case the profiles are set correctly in ApiInfoStore, but they never become visible, because we do not call `setInfo` anymore. For some reason this (only?) happens on iOS mobile devices.

There seem to be two ways two fix this: Either postpone the /info call until the state update callbacks are registered at the stores, i.e. until after this code: 
https://github.com/graphhopper/graphhopper-maps/blob/9a911c2e6caaba16f6bfb8466dbb934afbcfe43b/src/App.tsx#L54-L78

or explicitly call all the state update functions (`setInfo`, `setQuery`, ...) the first time the render function runs to make sure the stores' state is actually propagated to the app state.

When I tried this I ran into another problem: For some reason the above `useEffect` in App.tsx has no dependency list, so it runs on every render which resulted in an infinite loop of /info requests when I tried the first approach.  I don't know why this is, but I think there should be an empty dependency list to make sure the useEffect runs exactly once.

easbar:
Fixes #254.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -36,6 +36,7 @@ import useRoutingGraphLayer from '@/layers/UseRoutingGraphLayer'
 import MapFeaturePopup from '@/layers/MapFeaturePopup'
 import useUrbanDensityLayer from '@/layers/UseUrbanDensityLayer'
 import useMapBorderLayer from '@/layers/UseMapBorderLayer'
+import { getApi } from '@/api/Api'
 
 export const POPUP_CONTAINER_ID = 'popup-container'
 export const SIDEBAR_CONTENT_ID = 'sidebar-content'
@@ -67,6 +68,8 @@ export default function App() {
         getMapOptionsStore().register(onMapOptionsChanged)
         getPathDetailsStore().register(onPathDetailsChanged)
         getMapFeatureStore().register(onMapFeaturesChanged)
+
+        getApi().infoWithDispatch()
 
         return () => {
             getQueryStore().deregister(onQueryChanged)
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -68,6 +68,14 @@ export default function App() {
         getPathDetailsStore().register(onPathDetailsChanged)
         getMapFeatureStore().register(onMapFeaturesChanged)
 
+        onQueryChanged()
+        onInfoChanged()
+        onRouteChanged()
+        onErrorChanged()
+        onMapOptionsChanged()
+        onPathDetailsChanged()
+        onMapFeaturesChanged()
```
easbar, line 77: It can happen that the infoWithDispatch call returns **after** we initialized `setState` with the store state (so it is the initial inital state), but **before** these methods were registered for the stores above. Therefore we need to trigger the update explicitly here.

src/App.tsx
```diff
@@ -77,7 +85,7 @@ export default function App() {
             getPathDetailsStore().deregister(onPathDetailsChanged)
             getMapFeatureStore().deregister(onMapFeaturesChanged)
         }
-    })
+    }, [])
```
easbar, line 88: It made no sense that we ran this useEffect for every render so far?

PR #273, opened 2022-10-31
avoid appending to URL if default tile or default profile

### LINKED ISSUE #259: initial history should not be created for start page
Currently if you go to https://www.graphhopper.com/ then click "maps" the back button will only remove the profile parameter.

Instead it should go back to www.graphhopper.com/

Maybe we should not add parameters for the start page?

karussell:
This fixes #259

But it also avoids adding the default profile (and default layer) in general, not just on load. Not sure if this has some unwanted side-effects.


easbar:
Why do you want leave out the profile and the layer name in case they are the default? And what even is the default profile? Doesn't this depend on the server config? For example there could be no car profile at all.

PR #277, opened 2022-11-06
Do not update history initially and more

### LINKED ISSUE #259: initial history should not be created for start page

easbar:
Fixes #259.


src/NavBar.ts
```diff
@@ -11,21 +11,26 @@ import { getApi } from '@/api/Api'
...
-        this.queryStore.register(() => this.onQueryStateChanged())
         this.mapStore = mapStore
-        this.mapStore.register(() => this.onQueryStateChanged())
-        window.addEventListener('popstate', () => this.parseUrlAndReplaceQuery())
+        window.addEventListener('popstate', async () => await this.updateStateFromUrl())
+    }
+
+    async startSyncingUrlWithAppState() {
+        // our first history entry shall be the one that we end up with when the app loads for the first time
+        window.history.replaceState(null, '', this.createUrlFromState())
```
easbar, line 24: For example if we open 'http://localhost:3000' and due to the app state the URL is updated to 'http://localhost:3000/?whatever=here' clicking 'back' should lead us to where we came from, not just 'http://localhost:3000'. That's why we replace the current window history entry here instead of adding a new one.

src/NavBar.ts
```diff
@@ -11,21 +11,26 @@ import { getApi } from '@/api/Api'
...
+        this.queryStore.register(() => this.updateUrlFromState())
+        this.mapStore.register(() => this.updateUrlFromState())
```
easbar, line 26: Here we listen to updates of the query- and map store. This means whenever some user interaction updates both we would get two new history entries. This would probably not be what we want, but is a limitation of our current approach. 
To work around this we already have the 'ignoreStateUpdates' flag in 'createUrlFromState', because this is already the first case where we want to do multiple state updates for a single 'action'.
An alternative would be pushing history entries for certain user interactions explicitly instead of blindly reacting to these events here.

src/NavBar.ts
```diff
@@ -64,35 +69,6 @@ export default class NavBar {
...
-                        .geocode(p.queryText, 'nominatim')
-                        .then(res => {
-                            if (res.hits.length <= 0) return
-                            fullyInitPoints[idx] = {
-                                ...p,
-                                queryText: res.hits[0].name,
-                                coordinate: { lat: res.hits[0].point.lat, lng: res.hits[0].point.lng },
-                                isInitialized: true,
-                            }
-                            if (fullyInitPoints.every(p => p.isInitialized)) {
```
easbar, line 85: Here we got console errors sometimes because p could be undefined.

src/NavBar.ts
```diff
@@ -64,35 +69,6 @@ export default class NavBar {
...
-
-        // support legacy URLs without coordinates (not initialized) and only text, see #199
-        if (points.some(p => !p.isInitialized && p.queryText.length > 0)) {
-            if (!profile.name) profile = { name: 'car' }
-            let fullyInitPoints: QueryPoint[] = Array.from({ length: points.length })
-            points.forEach((p, idx) => {
-                if (p.isInitialized) fullyInitPoints[idx] = p
-                else
-                    getApi()
-                        .geocode(p.queryText, 'nominatim')
```
easbar, line 76: Here we got console errors in cases like `point=`, because apparently we get an error when geocoding with an empty string.

src/NavBar.ts
```diff
@@ -112,60 +88,78 @@ export default class NavBar {
...
-        for (let i = 0; i < result.length && i < parsedPoints.length; i++) {
-            result[i] = parsedPoints[i]
-        }
-
-        return result
+    private static dispatchQueryPoints(points: QueryPoint[]) {
+        // estimate map bounds from url points if there are any. this way we prevent loading tiles for the world view
+        // only to zoom to the route shortly after
+        const bbox = NavBar.getBBoxFromUrlPoints(points.filter(p => p.isInitialized).map(p => p.coordinate))
+        if (bbox) Dispatcher.dispatch(new SetInitialBBox(bbox))
```
easbar, line 144: We can do this for urls with `point=queryString' as well now, even though it doesn't help as much, because the (world) map already starts to load while the slow geocoding requests are still ongoing.

src/index.tsx
```diff
@@ -72,12 +72,16 @@ const smallScreenMediaQuery = window.matchMedia('(max-width: 44rem)')
...
 const navBar = new NavBar(getQueryStore(), getMapOptionsStore())
-// parse the initial url
-navBar.parseUrlAndReplaceQuery()
+
+// get infos about the api as soon as possible
+getApi()
+    .info()
+    .then(info => {
+        Dispatcher.dispatch(new InfoReceived(info))
+        navBar.updateStateFromUrl().then(() => navBar.startSyncingUrlWithAppState())
```
easbar, line 82: Since we are waiting for /info here we can be sure that the /info response does not trigger a second url update (and pushes another history entry). We also wait for `updateStateFromUrl` to return to make sure we do not get multiple history entries before and after the geocoding returns.

src/stores/QueryStore.ts
```diff
@@ -201,13 +201,24 @@ export default class QueryStore extends Store<QueryStoreState> {
                     queryText: queryText,
                 }
             })
+            // make sure there are always at least two input boxes
+            while (queryPoints.length < 2) {
```
easbar, line 205: Previously we handled this in NavBar where it was completely unrelated and did not even work for `point=queryString`.

src/NavBar.ts
```diff
@@ -64,35 +69,6 @@ export default class NavBar {
 
             return point
         })
-
-        // support legacy URLs without coordinates (not initialized) and only text, see #199
-        if (points.some(p => !p.isInitialized && p.queryText.length > 0)) {
-            if (!profile.name) profile = { name: 'car' }
```
easbar, line 70: This hard-coded 'car' profile was only used when there were point parameters with a query text and is removed now.

src/NavBar.ts
```diff
@@ -112,60 +88,78 @@ export default class NavBar {
...
+                            if (res.hits.length == 0) return p
+                            return {
+                                ...p,
+                                queryText: res.hits[0].name,
+                                coordinate: { lat: res.hits[0].point.lat, lng: res.hits[0].point.lng },
+                                isInitialized: true,
+                            }
+                        })
+                        // if the geocoding request fails we just keep the point as it is, just as if no results were found
+                        .catch(() => Promise.resolve(p))
```
easbar, line 125: We can also do something like http://localhost:3000/?point=&point=Berlin now which will resolve the second point from the 'Berlin' query string and the first will just be an empty input box.

PR #289, opened 2022-11-14
close sidebar

karussell:
For #20

If sidebar is closed path: should we use 100% of the width for path details widget? (but somehow avoid colliding with the attribution text after #283)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -144,13 +144,13 @@ interface LayoutProps {
 }
 
 function LargeScreenLayout({ query, route, map, error, mapOptions, info }: LayoutProps) {
-    let [visible, setVisible] = useState(true)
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -141,35 +144,47 @@ interface LayoutProps {
 }
 
 function LargeScreenLayout({ query, route, map, error, mapOptions, info }: LayoutProps) {
+    let [visible, setVisible] = useState(true)
```
easbar, line 147: const

easbar:
> should we use 100% of the width for path details widget? 

You mean the detail widget should expand/shrink dynamically when we open/close the sidebar? Sometimes the sidebar doesn't need so much vertical space so the path details could even be expanded when the sidebar is open. I'm not sure about this, did you try it yet?

PR #300, opened 2023-01-09
Consider profiles from config when /info returns

easbar:
This way the profile chosen initially should also be deterministic, at least when the profiles are set in the config.


easbar:
I opened https://graphhopper.com/maps/ several times, but the profile was always 'car'. Or how can you reproduce this?

easbar:
Ah, this depends on our backend. Anyway, this should be fixed by 76f471d6c3336e112527d0d4bb433e17e33f1212

PR #308, opened 2023-02-13
Dark mode

karussell:
Dark mode is IMO in general not worth the effort, but very useful for driving at night (#279).

In Firefox you can explicitly switch to dark mode in the settings and then you can try it [here](https://graphhopper.com/maps-dev/dark_mode/?profile=car&layer=Mapilion+Dark).

Still there are many TODOs:

 * make it possible to switch via javascript. Use something like [this](https://stackoverflow.com/a/58543220/194609) which should avoid the media queries all over the code too.
 * didn't find a way to change the CSS for custom model box and heightgraph -> this was probably fixed in #318
 * white graphhopper logo for dark mode
 * buy stadia maps standard and remove maptiler or fix maptiler for dark mode
 * a "mouse over" over the profile icons should not nearly hide them
 * white is too bright for auto complete and preview panel background is still white
 * set black background for selectors that set padding otherwise it is white frame like the map options or context menu

grafik

### DIFF
--- a/src/custom.d.ts
+++ b/src/custom.d.ts
@@ -11,7 +11,7 @@ declare module 'config' {
     const keys: {
         graphhopper: string
         omniscale: string
-        maptiler: string
+        stadiamaps: string
         thunderforest: string
         kurviger: string
     }
--- a/src/index.html
+++ b/src/index.html
@@ -29,6 +29,21 @@
                 min-height: 100%;
             }
 
+            /* TODO NOW:
+               1. set black background for selectors that set padding otherwise it is white frame like the map options or context menu
+               2. white graphhopper logo for dark mode
+               3. custom model box and heightgraph
+               4. buy stadia maps standard and remove maptiler or fix maptiler for dark mode
+               5. over over icons should not nearly hide them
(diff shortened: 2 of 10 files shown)


Janekdererste:
Still getting notifications on this project :-)

The following is pretty similar to the stackoverflow link, but slightly different I think. You can set css variables within media queries too. I found this really helpfull when I implemented a dark mode. 
```
:root {
            --header: #04A777;
            --highlight: #C92C6D;
            --text: #011627;
            --background: #fff;
        }

        @media (prefers-color-scheme: dark) {
            :root {
                --header: #04A777;
                --highlight: #C92C6D;
                --text: #FAFAFA;
                --background: #232327;
            }
        }
``` 
and then use those vars throughout the css
```
html, body {
            margin: 0;
            padding: 0;
            height: 100%;
            background: var(--background);
            font: 100% system-ui;
        }

        a, body {
            color: var(--text)
        }
```

Janekdererste:
You could do something along the lines of these snippets https://css-tricks.com/updating-a-css-variable-with-javascript/

PR #316, opened 2023-03-16
Allow custom model for mobile devices (plus separate settings panel)

karussell:
In master the custom model is only active if the custom model box is visible which leads to bugs: 
1. We disabled custom model for mobile devices (small screens) as the app would become unusable with an open custom model box. And so links with a custom model in the URL will (silently) show the wrong route on mobile devices as the custom model is removed. 
2. switching between mobile and desktop sizes does not work as the custom model gets lost
3. custom model gets lost or encoded values are not properly initialized when minimizing (with the x button) and then showing the sidebar again (hamburger button)

In this PR changes I basically moved the custom-model object state from the custom-model-editor into the SettingsStore that lives regardless of the UI component.

And so we can change how the gear wheel button works: the button now just opens a new settings box where we can change the miles/kilometer settings and also if the custom model is enabled or not. (A small shadow under the gear wheel is shown when the custom model is enabled. I'm open to suggestions to make it more obvious.)

And because the custom model editor does no longer store the custom model itself and also not if the custom model is enabled we don't have to always show the custom model editor and the above problems for small screen sizes disappear and we can additionally support custom models for mobile devices. Furthermore we can easily add new settings (like the dark mode option) without adding more stuff to the main view.

 - [x] trigger routing if "custom_model" in URL parameter (but keep settings box closed). For that I had to use and extract the validateJson method from the custom model editor.
 - [x] avoid route re-calculation if settings box is shown (with enabled custom model box)
 - [x] ~make json invalid -> hide settings box -> now default custom model example is shown instead of invalid (completely different) custom model before -> for now fixed it and closing the settings box means "reset to custom model before invalidated one". Is different than before but IMO makes sense now.~
 - [x] fix [a tricky problem](https://github.com/graphhopper/graphhopper-maps/pull/180#issuecomment-1475392948) with jest and ES6 modules -> no longer necessary
 - [x] remove QueryStore.state.customModel as we have it in the SettingsStore

You can try it out [here](https://graphhopper.com/maps-dev/gear_as_setting_button/?point=51.175065%2C11.980286&point=51.257645%2C12.655945&profile=car&layer=Omniscale&custom_model=%7B%22priority%22%3A%5B%7B%22if%22%3A%22road_class+%3D%3D+MOTORWAY%22%2C%22multiply_by%22%3A%220.0%22%7D%5D%7D).

grafik

### DIFF
--- a/src/sidebar/SettingsBox.tsx
+++ b/src/sidebar/SettingsBox.tsx
@@ -1,29 +1,26 @@
-import {ToggleDistanceUnits} from '@/actions/Actions'
+import { ToggleDistanceUnits } from '@/actions/Actions'
 import Dispatcher from '@/stores/Dispatcher'
 import styles from '@/sidebar/SettingsBox.module.css'
-import {tr} from '@/translation/Translation'
+import { tr } from '@/translation/Translation'
 import PlainButton from '@/PlainButton'
 import OnIcon from '@/sidebar/toggle_on.svg'
 import OffIcon from '@/sidebar/toggle_off.svg'
-import {useContext} from 'react'
-import {ShowDistanceInMilesContext} from '@/ShowDistanceInMilesContext'
+import { useContext } from 'react'
+import { ShowDistanceInMilesContext } from '@/ShowDistanceInMilesContext'
 
 export default function SettingsBox() {
     const showDistanceInMiles = useContext(ShowDistanceInMilesContext)
     return (
-
         <div className={styles.parent}>
-            <div className={styles.title}>
-                {tr('settings')}
-            </div>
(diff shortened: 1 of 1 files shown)


easbar:
> switching between mobile and desktop sizes does not work as the custom model gets lost
custom model gets lost or encoded values are not properly initialized when minimizing (with the x button) and then showing the sidebar again (hamburger button)

This is a little bug we introduce when we added the custom model examples: We overwrite the current custom model with the default example, even when we entered our own model before. It can be fixed like this:

https://github.com/graphhopper/graphhopper-maps/commit/b2263608d98fddd9180f0e42fefffba87227f01c

PR #318, opened 2023-03-21
use css for custom model box from local file

karussell:
This change makes it possible to change/overwrite the editor css from here.


webpack.common.js
```diff
@@ -57,7 +57,9 @@ module.exports = {
                     {
                         loader: 'css-loader',
                         options: {
-                            modules: true,
+                            modules: {
+                                auto: (resourcePath) => resourcePath.endsWith(".module.css")
+                            },
```
easbar, line 62: This means we apply this rule to **all** css files (see the test property) (except node_modules), but use the modules only for those files ending with `.module.css`. There are probably other ways to achieve the same, but at least we can use plain css where 'needed' now. We 'need' this now, because we need to control the class names for the codemirror css rules and our current solution is to not use css modules for these css files.

easbar (approved):

PR #322, opened 2023-03-27
detailed error message

oblonski:


src/api/Api.ts
```diff
@@ -62,13 +62,17 @@ export class ApiImpl implements Api {
     async info(): Promise<ApiInfo> {
         const response = await fetch(this.getRoutingURLWithKey('info').toString(), {
             headers: { Accept: 'application/json' },
-        })
+        }).catch(() => { throw new Error('Could not connect to the Service. Try to reload!') });
```
oblonski, line 65: Any network issues where requests cannot be sent or responses cannot be received should be catched here, and "Could not connect to the Service" pretty much sums this up.

src/api/Api.ts
```diff
@@ -62,13 +62,17 @@ export class ApiImpl implements Api {
...
 
+        const result = await response.json()
```
oblonski, line 67: If we are here, we can use what the server says.

oblonski:
Ah already deployed. Nice error for:

`https://graphhopper.com/maps/?point=48.123299%2C11.543011_GraphHopper+GmbH%2C+Lipowskystra%C3%9Fe+30%2C+81373+Munich%2C+Germany&point=48.118435%2C11.539495_Lindenschmitstra%C3%9Fe+52%2C+81373+Munich%2C+Germany&profile=car&layer=Omniscale&key=bloed`

PR #327, opened 2023-03-28
visualize network access, i.e. queried point to snapped waypoint

oblonski:
looks like this

Screenshot 2023-03-28 at 20 18 27


oblonski:
Screenshot 2023-03-28 at 20 20 35

src/layers/UsePathsLayer.tsx
```diff
@@ -73,6 +78,58 @@ function addUnselectedPathsLayer(map: Map, paths: Path[]) {
     map.addInteraction(select)
 }
 
+function createBezierLineString(start: number[], end: number[]):LineString {
+    var curvePoints = [];
```
easbar, line 82: You can always use `const`, unless you get a (Typescript) compiler error. Then you can use `let`. `var` is basically as thing of the past.

src/layers/UsePathsLayer.tsx
```diff
@@ -73,6 +78,58 @@ function addUnselectedPathsLayer(map: Map, paths: Path[]) {
...
+            lineDash: [1, 10],
+            lineCap: 'round',
+            lineJoin: 'round'
+        }),
+    })
+    var layer = new VectorLayer({
+        source: new VectorSource()
+    });
+    layer.setStyle(style);
+    for(let i=0;i<selectedPath.snapped_waypoints.coordinates.length;i++){
```
easbar, line 123: For consistent formatting you can run `npm run format` or integrate the prettier configuration with your IDE or text editor.

easbar:
> looks like this

Looks very nice :)

src/layers/UsePathsLayer.tsx
oblonski: done

src/layers/UsePathsLayer.tsx
oblonski: Oh ja, thanks for the hint. Just did it.

easbar: Du kannst das natürlich auch mergen, oder soll ich?

oblonski: Mach lieber du. Ich bin hier nur Gast.

oblonski: Danke @easbar

easbar: Ein willkommener Gast. Danke dir!

Issue #282, opened 2022-11-12
Zoom to current location button


easbar:
Or even something like 'track current location' and show a pin for the current position.

PR #339, opened 2023-06-04
Current location button

karussell:
For #282. 

TODOs:

- [x] set from location and marker if undefined "from"-input. (Or even overwrite existing "from"-input? Google maps web just zooms to current location.)
- [x] fix css glitch when opening the map layer

[Try here](https://graphhopper.com/maps-dev/curr_loc).

grafik

### DIFF
--- a/src/App.module.css
+++ b/src/App.module.css
@@ -54,28 +54,22 @@
 }
 
 .onMapRightSide {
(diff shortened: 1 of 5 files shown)


easbar:
Works for me and quite useful IMO :+1:

PR #341, opened 2023-06-09
avoid response interference for slow queries

karussell:
If a custom model is enabled (or if no CH is activated on the server in general) it happens that an earlier sent request comes back later. For this simple query interference there was already a protection in place but in the following scenario it does not work:

 * send query A (takes 10s)
 * after 1s send query B (takes 1s)
 * after 1s response from B comes back and is shown on the map
 * after 1s send query C (takes 10s or at least so long that query B already respond)
 * after 7s query A comes back and the route is shown on the map although the markers are already prepared for query C (this is the bug!)
 * after 3s query C comes back

In these considerations alternatives routes make everything a bit complicated but they were not the underlying reason of the problem.

### DIFF
--- a/src/api/Api.ts
+++ b/src/api/Api.ts
@@ -163,8 +163,8 @@ export class ApiImpl implements Api {
                 }
             })
             .catch(error => {
-                console.warn('error when performing /route request ' + routeNumber + ': ', error)
                 if (routeNumber > this.lastRouteNumber) {
+                    console.warn('error when performing /route request ' + routeNumber + ': ', error)
                     this.lastRouteNumber = routeNumber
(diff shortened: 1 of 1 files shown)


src/api/Api.ts
```diff
@@ -53,6 +52,8 @@ export class ApiImpl implements Api {
     private readonly apiKey: string
     private readonly routingApi: string
     private readonly geocodingApi: string
+    private routeCounter = 0
+    private lastRouteNumber = -1
```
easbar, line 56: maybe stick to either counter or number?

easbar: ok :)

PR #379, opened 2024-04-09
Better profile icons

### LINKED ISSUE #376: Change icon for unknown profiles?
In https://graphhopper.slack.com/archives/DCYDKS5RB/p1710765026728319 we added a question mark icon for unknown profiles. The UI might look like this for custom profile names:

image

This looks somewhat unintentional and buggy. The good thing is that we get to see the profile name when we hover the icons. But maybe we can use another icon in this case?

SachsenspieltCoding:
fixes #376 

# Better profile icons

This PR changes the behaviour of the profile icons, which is an addition to #368.

## Icons can be added simpler

```ts
// ALL AVAILABLE ICONS
// every svg gets mapped to a key, so icons can be easily added
export const icons = {
    car: CarIcon,
    small_truck: SmallTruckIcon,
    truck: TruckIcon,
    scooter: ScooterIcon,
    foot: FootIcon,
    hike: HikeIcon,
    bike: BicycleIcon,
    mtb: MtbBicycleIcon, // Mountainbike
    racingbike: RacingbikeIcon,
    motorcycle: MotorcycleIcon,
    wheelchair: WheelchairIcon,
    question_mark: QuestionMarkIcon,
}
```

From now on, the icons are defined with a map inside the `profileIcons.ts`. Every available SVG get's mapped to a specific key, so it will automatically be used, without any further code changes.

## Replaced question marks with numbers

image

When there is no icon available for that profile, an incremental number will be displayed.

## Icons based on prefix

Icons now get their icons, based on their prefix. While standard profiles like `car` or `foot` still get their default icon, every profile with one of those as a prefix also gets the same base icon, but with a small number in the top right corner. Therefore, icons for custom profiles are now supported in a way. 

**Example**
image

`... car_delivery, car_avoid_ferry, car_avoid_motorway, car_avoid_toll, small_truck_delivery, ...`


SachsenspieltCoding:
Oh, that is a bug, right.
This could easily be fixed if, as you said, we did not include car itself in the list of profiles with car as a prefix, as it is ignored anyway, so is obsolete.

SachsenspieltCoding:
@karussell Should be fixed now

SachsenspieltCoding:
grafik

Issue #385, opened 2024-05-18
QOL: Sort autocomplete entries by distance


SachsenspieltCoding:
When the current location is selected as starting point, the autocomplete items for the waypoints should be sorted by (air) distance

PR #390, opened 2024-05-22
QOL: Autocomplete entries sorted by distance

SachsenspieltCoding:
Since I found it quite annoying that the autocompletion was always based on the coordinate `0,0`, I implemented that if there is a query point above the currently selected one, these coordinates are sent together with the geocode request to the GraphHopper Geocoding API.

### DIFF
--- a/src/api/Api.ts
+++ b/src/api/Api.ts
@@ -28,7 +28,7 @@ export default interface Api {
 
     routeWithDispatch(args: RoutingArgs, zoom: boolean): void
 
-    geocode(query: string, provider: string): Promise<GeocodingResult>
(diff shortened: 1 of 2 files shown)


SachsenspieltCoding:
fixes #385

SachsenspieltCoding:
Ah yes, I admit that this realisation follows the motto "quick'n'dirty"

I'll look into this :)

SachsenspieltCoding:
This should do it

SachsenspieltCoding:
I've moved the logic part out of the Geocode class, could you check if this is what you had in mind?

Issue #419, opened 2025-04-28
Map Options don't disappear after tapping the outside area on mobile


avmvng:
On mobile, when the layers button is tapped and Map Options are shown, the only way to make them go away is to select a different layer or re-select the current layer (most likely OpenStreetMaps) and then tap outside the Map Options menu.

__Expected behavior:__

After the Map Options menu is opened, the user should be able to tap anywhere on the map to make it disappear without having to select or re-select a layer option.

__Current behavior:__

The user has to select a new layer (Omniscale, Cyclosm, etc.) in order to close the Map Options menu. There are two cases:

1. The user _actually_ wants to switch to a different map. Then the user selects it, which updates the page and closes the menu.
2. The user does not want to switch to a different map. Then, tapping outside the menu area does not work without selecting the current map and THEN tapping outside the menu area.

__Steps to reproduce:__
1. Open Graphhopper Maps on mobile
2. Tap on the layers button
3. Tap outside of the layer selection area

Attaching a video demonstrating the issue:

https://github.com/user-attachments/assets/53dd0dcc-e53a-4598-ac0a-c71b04631a3e

I'm happy to work on this issue so let me know if it's worth looking into!

PR #420, opened 2025-05-01
MapOptions close by clicking outside component on mobile

avmvng:
This PR addresses #419 by using `useEffect` and `useRef` in the `MapOptions` component so it closes automatically when a user clicks outside of it (this is an issue only on mobile, the desktop functionality is unchanged).

### DIFF
--- a/src/map/MapOptions.tsx
+++ b/src/map/MapOptions.tsx
@@ -1,4 +1,4 @@
-import { useState } from 'react'
+import { useState, useEffect, useRef } from 'react'
 import styles from './MapOptions.module.css'
 import { MapOptionsStoreState } from '@/stores/MapOptionsStore'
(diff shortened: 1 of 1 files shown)


avmvng:
@karussell Great question! On Desktop, you can only choose between different layers by hovering over the layers button, thanks to this logic:
```
const [isOpen, setIsOpen] = useState(false)
/* other code */
onMouseEnter={() => setIsOpen(true)}
onMouseLeave={() => setIsOpen(false)}
```
On Mobile, the only way to open this dropdown component is clicking the button, and the only way to close it is to trigger `notifyChanged={() => setIsOpen(false)}`. Thus this is not a React/OpenLayers bug but rather just a UI thing that wasn't implemented previously.

Issue #334, opened 2023-04-22
Customizable line properties of suggested routes


apophenic-treehugger:
I suggest a visual improvement to the online routing interface.

**Problem**
When using the newly introduced cyclosm layer in the online router (#317 ), there are two related problems:
-  It can be difficult to distinguish the route suggested by graphhopper from the blue cycle infrastructure of the basemap.
- The suggested route obscures valuable information of the basemap below ("Is the thing below the route a cycle path or a road?")

**Example**
[Here](https://graphhopper.com/maps/?point=44.840553%2C-0.581155&point=45.001084%2C-1.201221&profile=bike&layer=Cyclosm) it is hard or impossible to tell whether the route leads along a cycle track or a road.

**Suggested solution**
Add a few simple slidebar controls to set the line properties:
- colour
- transparency
- width
- style (solid, dashed, dotted)

The user can then change them as they deem fit.

_Additionally to my suggestion, it may make sense to already set the defaults to different values when cyclosm is used as baselayer._

ratrun:
A probably simpler solution is to define a key with which it is possible to hide the route as long as the key is pressed. Komoot uses the `m' key for this.

apophenic-treehugger:
`m` like **m**oooooooore base**m**ap!

easbar:
If someone has a suggestion how to improve the current color/style/width we could also change it instead of adding controls.

apophenic-treehugger:
I think it's good if the line is dashed.

How about a thin dashed yellow line (somewhere around #fff176) with a semi-transparent black halo (#000000a0 or so) ?

apophenic-treehugger:
So what do you propose?

apophenic-treehugger:
I'm fine with whatever solution that solves the problem. Obviously, there are various ways to do this. I've tried to make constructive suggestions and they were rejected. I don't feel like I have the authority to decide this, nor do I have the skills to implement any of the solutions discussed.

ratrun:
I'm very sorry for all the hassle. Unfortunately I'm not really familiar with react 😢. But I was able to reproduce it and I think the problem was that the map element didn't receive the focus. As the PR #428 was already closed I created a new PR: #435

PR #428, opened 2025-06-07
Hide paths while pressing the "h" key or press a small icon

ratrun:
Implement the suggested solution for #334. While pressing the "h" key, the paths are removed such that one can see the map details of the roads. Additionally, a small button next right to the map attribution also hides or shows again the paths.

Please review and provide feedback if there is a chance that this gets merged. Currently, the translation for the button tool tip is missing as I didn't want to change the spreadsheet as I don't know if this can get merged.

### DIFF
--- a/src/layers/UsePathsLayer.tsx
+++ b/src/layers/UsePathsLayer.tsx
@@ -24,16 +24,18 @@ export default function usePathsLayer(map: Map, paths: Path[], selectedPath: Pat
     const [showPaths, setShowPaths] = useState(true)
     useEffect(() => {
         removeCurrentPathLayers(map)
-        addUnselectedPathsLayer(
-            map,
-            paths.filter(p => p != selectedPath)
-        )
-        addSelectedPathsLayer(map, selectedPath)
-        addAccessNetworkLayer(map, selectedPath, queryPoints)
+        if (showPaths) {
+            addUnselectedPathsLayer(
+                map,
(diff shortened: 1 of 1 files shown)


ratrun:
I just detected that the button does not appear in the small screen and its absolute positioning is problematic with other attribution lengths. It only works with the Openstreetmap map currently.

ratrun:
> Why is a button necessary? We could document it in keyboard shortcuts or something.

Without an explicit button (which indicates the respective shortcut key) nobody will figure out that this feature exists. And who reads a documentation about shortcuts? I didn't even know that such documentation exists :smiling_face_with_tear:

ratrun:
In the meantime I moved the button to the left of the attribution. Please check!

ratrun:
> There a few changes that are unnecessary if you move the logic into usePathsLayer where the path could be conditionally added/removed directly. Or was there a special reason you duplicated the code into the MapComponent?
> 
> And other changes are still related to the button. Can you remove it?

Oh, sorry! I pushed the wrong version and didn't notice. Now it should be OK.

apophenic-treehugger:
Sorry for reopening this but I hadn't been notified about this discussion since it took place outside the #334 thread.

Thanks a lot for implementing this!

However, I agree with @ratrun that without a button on the UI, nobody except the three of us will notice and hence use this feature.

PR #449, opened 2026-03-23
Favourite locations

karussell:
An attempt to automatically save recent locations to avoid typing. Will be shown only if a location was successfully used in a route for more than once and only if 0 or 1 character is in the input box (to avoid mixing it with the geocoding results). Also this can be easily cleared.

Try it [here](https://graphhopper.com/maps-dev/favourite_locations/).

image


karussell:
Will merge but let me know if there is some obvious ugliness I'm unable to see ;)

Janekdererste:
Sorry, I was on vacation. The clock on the left is a good choice :-)

PR #451, opened 2026-04-29
Use Cmd+Enter as keyboard shortcut in custom model editor for macOS

easbar:
* Ctrl+Enter conflicts with macOS context menu shortcut
* apparently macOS suppresses keyup if Cmd is pressed


karussell:
Thanks!

PR #462, opened 2026-08-21
Add reverse route button

ratrun:
I added the possibility to reverse the route right to the settings button:
grafik
The change was created with the help of AI.


karussell:
Thanks!

I do not like this position for a feature like this as it is already very crowded at this position. 

What about the context menu of the map? (we could remove "center map" while we are at it).

ratrun:
> What about the context menu of the map? (we could remove "center map" while we are at it).

I fear that nobody will expect this functionality there. It does not seem to be the right position. I checked how Google and komoot do this. The Google approach does not really work for us because we support via points. Komoot uses a similar position as compared to my PR. But they have a dropdown list for the available profiles, which seems a better concept to me.

The problem is that I don't find another logical position.

ratrun:
Maybe here where I put the "x"?

grafik

karussell:
Aren't there other route-modifying functions that could be equally added like "create round trip"?

IMO it is rather easy to add new functions and distract the average user and we should say 99% of the time "no" to new buttons :)

ratrun:
I just implemented your suggestion and also removed the center map functionality. Please check.

karussell:
Thanks, looks good. Two nitpicks: the formatting seems to be off (just do `npm run format`)

And can you also update the [translation spreadsheet](https://docs.google.com/spreadsheets/d/18z00Rbt6QvLIkayEV9P89vW9oU0QbTVsjRk9nz1CeFY/edit?gid=0#gid=0) directly?

ratrun:
I ran `npm run format` and updated the spreadsheet.

karussell:
Thanks!

PR #463, opened 2026-09-04
feat(lang): finish french language

ywaymel:
I finished the missing translations for french (fr_FR).


ywaymel:
I need a confirmation about POI entries: in french, a ton of them are most naturally written with three words (although it's specified in C212 that there should be at most two).
Usually, it's just a preposition added in the middle (like "de" or "aux"), so nothing major. Is this rule here for technical reasons or is it just to avoid having too specific entries (e.g., extra adjectives)?

karussell:
Thanks! Three world phrases are fine now. Please rebase or merge master into here.

> the two-word limit in C212 is just a note in the translation sheet, not a code limit.

Have fixed this comment in the translation sheet.

Also add "à" to poi_removal_words, so "… à lyon" doesn't keep "à" in the location.

PR #466, opened 2026-09-22
Center map on request points if no route could be found

### LINKED ISSUE #306: Map doesn't center if no route could be found
See https://github.com/graphhopper/graphhopper/issues/2745

https://graphhopper.com/maps/?point=68.95933%2C179.993481&point=68.954544%2C-179.954919&profile=car&layer=Omniscale

## Expected behaviour

The map should be centered at the selected start/end points even if no route could be found

image

## Actual behaviour

image

jp-roisin:
Fixes #306

When a route request fails (e.g. "Connection between locations not found") the map now fits the view to the requested points instead of not moving at all.

Changes:
- `MapActionReceiver` handles `RouteRequestFailed` and fits the map to the request points, reusing the existing `getBBoxPoints` helper.
- `getBBoxPoints` wraps around the antimeridian when the points are closer to each other that way, so the example from #306 (points on either side of ±180°) zooms in on the points instead of showing the whole world. This also fixes the initial view when loading such a URL.
- added a small unit test for `getBBoxPoints`.

Verified manually with the URL from the issue (`?point=68.95933,179.993481&point=68.954544,-179.954919&profile=car`): the map now centers across the antimeridian with both markers visible, matching the expected screenshot in the issue. Also verified a regular failed request (Longyearbyen → Tromsø by car) fits both points, and that successful routing is unchanged.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


karussell:
Thanks, this works!

